PR #4, opened 2020-05-18
Add label of PowerShell

kenchan0130:
[PowerShell](https://github.com/powershell/powershell) is often used when working with Office 365 or Azure AD.

Main Fix:
https://github.com/scriptingosx/Installomator/commit/8ec2179c1c68b179e080fa12197db1e5f4cf4b44

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -45,7 +45,7 @@ BLOCKING_PROCESS_ACTION=prompt_user
 #     - zip
 #     - pkgInDmg
 #     - pkgInZip (not yet tested)
-# 
+#
 # - downloadURL: (required)
 #   URL to download the dmg.
 #   Can be generated with a series of commands (see BBEdit for an example).
@@ -61,7 +61,7 @@ BLOCKING_PROCESS_ACTION=prompt_user
 #     spctl -a -vv -t install ~/Downloads/desktoppr-0.2.pkg
 #
 #   The team ID is the ten-digit ID at the end of the line starting with 'origin='
-# 
+#
 # - archiveName: (optional)
 #   The name of the downloaded file.
 #   When not given the archiveName is derived from the $name.
@@ -74,7 +74,7 @@ BLOCKING_PROCESS_ACTION=prompt_user
 #   dmg or zip:
 #     Applications will be copied to this directory.
 #     Default value is '/Applications' for dmg and zip installations.
-#   pkg: 
+#   pkg:
 #     targetDir is used as the install-location. Default is '/'.
 #
 # - blockingProcesses: (optional)
@@ -85,7 +85,7 @@ BLOCKING_PROCESS_ACTION=prompt_user
 #     blockingProcesses=( "Keynote" "Pages" "Numbers" )
 #   When a workflow has no blocking processes, use
 #     blockingProcesses=( NONE )
-# 
+#
 # - pkgName: (optional, only used for dmgInPkg and dmgInZip)
 #   File name of the pkg file _inside_ the dmg or zip
 #   When not given the pkgName is derived from the $name
@@ -118,7 +118,7 @@ BLOCKING_PROCESS_ACTION=prompt_user
 downloadURLFromGit() { # $1 git user name, $2 git repo name
     gitusername=${1?:"no git user name"}
     gitreponame=${2?:"no git repo name"}
-    
+
     if [ -n "$archiveName" ]; then
     downloadURL=$(curl --silent --fail "https://api.github.com/repos/$gitusername/$gitreponame/releases/latest" \
     | awk -F '"' "/browser_download_url/ && /$archiveName/ { print \$4 }")
@@ -166,7 +166,7 @@ case $label in
         echo "Installomater: version $VERSION ($VERSIONDATE)"
         exit 0
         ;;
-    
+
     # label descriptions start here
     googlechrome)
         name="Google Chrome"
@@ -289,7 +289,7 @@ case $label in
             | cut -d '"' -f 2 )
         expectedTeamID="UPXU4CQZ5P"
         ;;
-    boxdrive) 
+    boxdrive)
         # credit: Isaac Ordonez, Mann consulting
         name="Box"
         type="pkg"
@@ -308,7 +308,7 @@ case $label in
         name="Zoom.us"
         type="pkg"
         downloadURL="https://zoom.us/client/latest/ZoomInstallerIT.pkg"
-        expectedTeamID="BJ4HAAB9B3"        
+        expectedTeamID="BJ4HAAB9B3"
         blockingProcesses=( zoom.us )
         ;;
     sonos)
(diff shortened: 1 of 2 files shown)


Installomator.sh
```diff
@@ -635,9 +635,14 @@ case $label in
         updateTool="/Library/Application Support/Microsoft/MAU2.0/Microsoft AutoUpdate.app/Contents/MacOS/msupdate"
         updateToolArguments=( --install --apps WDAV00 )
         ;;
+    powershell)
+        name="PowerShell"
+        type="pkg"
+        downloadURL=$(downloadURLFromGit PowerShell PowerShell | grep lts)
```
kenchan0130, line 641: The PowerShell project has 2 packages for macOS.
I chose lts package because I thought lts was more appropriate.

scriptingosx:
The label as it fails to retrieve the download url. Since powershell has such a large number of available downloads for each release the `downURLFromGit` function is not useful here.

You'll have to find a way to custom parse the gitHub output (see `handbrake` label for an example)

kenchan0130:
> The label as it fails to retrieve the download url. Since powershell has such a large number of available downloads for each release the downURLFromGit function is not useful here.

I am aware of this. For that reason, I use the grep command to uniquely target the package.

```sh
downloadURLFromGit PowerShell PowerShell | grep lts
```

If this method is not recommended, I  will try do it like `handbrake` without using a function .

scriptingosx:
the `downloadURLFromGit` will fail on this repository, so the `grep` doesn't do anything

kenchan0130:
I tried to do debug with the following case statement.

```sh
powershell)
        name="PowerShell"
        type="pkg"
        echo "Debug powershell."
        downloadURL=$(downloadURLFromGit PowerShell PowerShell | grep lts)
        echo ${downloadURL}
        exit 0
        expectedTeamID="UBF8T346G9"
        ;;
```

```sh
$ ./Installomator.sh powershell
Debug powershell.
https://github.com/PowerShell/PowerShell/releases/download/v7.0.1/powershell-lts-7.0.1-osx-x64.pkg
```

It seems that there is no particular problem.

---

Debug environment

```sh
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G2022
```

kenchan0130:
I may have identified the cause of this problem.
It's probably the rate limit of the GitHub API.

```sh
$ curl -fsS "https://api.github.com/repos/powershell/powershell/releases/latest"
curl: (22) The requested URL returned error: 403 rate limit exceeded
```

We can relax the rate limit by using GitHub's API token.

```sh
$ curl -fsS -H "authorization: Bearer ${GITHUB_API_TOKEN}" "https://api.github.com/repos/powershell/powershell/releases/latest"
```

scriptingosx:
No, still failing for me: 

```
$ sudo Desktop/Installomator.sh powershell
Start debug powershell

End debug powershell
```

I tested both on 10.14.6 and 10.15. And other GitHub applications are working, so it's not rate limited

kenchan0130:
Maybe it depends on me or your environment.

```sh
curl -fsS "https://api.github.com/repos/powershell/powershell/releases/latest" | awk -F '"' '/browser_download_url/ && /pkg/ { print $4 }'
```

This script is the actual processing of `downloadURLFromGit` function.

In my environment,

```sh
sh-3.2$ /usr/bin/curl -fsS "https://api.github.com/repos/powershell/powershell/releases/latest" | /usr/bin/awk -F '"' '/browser_download_url/ && /pkg/ { print $4 }'
https://github.com/PowerShell/PowerShell/releases/download/v7.0.1/powershell-7.0.1-osx-x64.pkg
https://github.com/PowerShell/PowerShell/releases/download/v7.0.1/powershell-lts-7.0.1-osx-x64.pkg
```

Two URLs are output.

Please check that multiple urls are output below to help isolate the problem.

scriptingosx:
I found a way to make it work. I added two labels: `powershell` and `powershell-lts` in case someone needs the other one.

kenchan0130:
@scriptingosx 
It may have been temporarily out of notarization.
I have now confirmed that it is notarized with no problems.

I would appreciate it if you could remove the comment out.

scriptingosx:
I still see it as signed, but not Notarized:

```
% spctl -a -vv -t install PowerShell.pkg
PowerShell.pkg: rejected
source=Unnotarized Developer ID
origin=Developer ID Installer: Microsoft Corporation (UBF8T346G9)
```

kenchan0130:
In my environment, they appear to be notarized.

```sh
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G6020
```

```sh
$ shasum -a 256 powershell-7.0.3-osx-x64.pkg
2c7702956f11b037cf75bf78372a9b0cb34d102031c0a293342de21c741800ed  powershell-7.0.3-osx-x64.pkg
$ spctl -a -vv -t install powershell-7.0.3-osx-x64.pkg
powershell-7.0.3-osx-x64.pkg: accepted
source=Developer ID
origin=Developer ID Installer: Microsoft Corporation (UBF8T346G9)
```

```sh
$ shasum -a 256 powershell-lts-7.0.3-osx-x64.pkg
e780050b75d21e0e37d9c589e423409a6f84cb74c67c177566979d63a75ede23  powershell-lts-7.0.3-osx-x64.pkg
$ spctl -a -vv -t install powershell-lts-7.0.3-osx-x64.pkg
powershell-lts-7.0.3-osx-x64.pkg: accepted
source=Developer ID
origin=Developer ID Installer: Microsoft Corporation (UBF8T346G9)
```

scriptingosx:
That is _signed_ not _notarized. When it is notarized, it should say `source=Notarized Developer ID`

kenchan0130:
I understood about that.

kenchan0130:
I recognize that until https://github.com/PowerShell/PowerShell/issues/10874 is resolved, introduction will be difficult.

scriptingosx:
I had been thinking about making notarization optional #65

Theile:
Maybe devastating, but what if the TeamID variable is not set, and the script would not check anything like this? I understand the implications, but in Denmark some schools need WordMat from GitHub, and that is not even signed…

scriptingosx:
My reasoning here is the following: when an app is not signed at all is is the admin's duty to verify the download and make sure the correct version is deployed to the client. For example, it is not that hard to hijack a Github account and create a new version that contains malware, spyware or adware.

So, you _should_ manually download the app in question, verify that it is untampered with and not infected with malware, repackage and upload to the management system.

Installomator uses the signature to verify that app has not been tampered with since signing and compares team IDs to check if the signer was the right entity. In this case a malicious attacker would have to compromise the hosting server _and_ the Developer Team ID certificates. While this is conceivable, such a breach would fool _every other_ security mechanism in macOS as well.

When an app is unsigned, Installomator cannot verify _anything_. If someone were able to hijack DNS resolution in your network, they could redirect requests to their software. I would not be comfortable with a deployment like that.

An signed but not-notarized app, such as powershell or Packages.app, is a different issue. In this case we can still verify that the archive has not been tampered with since signing and compare team IDs. All that we are missing is the additional approval from Apple that they scanned it during Notarization (whatever they scan it there for).

That's why I have been considering making the notarization check optional  ( #65 ). So far, there was only one piece of software that was affected this way (Whitebox Packages.app) so it didn't seem worth the effort. Powershell is the second I have come across.

PR #12, opened 2020-05-20
Add Docker Desktop & Brave Browser

securitygeneration:
Add Docker Desktop

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -467,7 +467,19 @@ case $label in
         downloadURL=$(curl -fs https://royaltsx-v4.royalapps.com/updates_stable | xpath '//rss/channel/item[1]/enclosure/@url'  2>/dev/null | cut -d '"' -f 2)
         expectedTeamID="VXP8K9EDP6"
         ;;
-        
+    docker)
+        name="Docker"
+        type="dmg"
+        downloadURL="https://download.docker.com/mac/stable/Docker.dmg"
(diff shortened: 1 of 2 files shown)


Installomator.sh
```diff
@@ -467,7 +467,19 @@ case $label in
         downloadURL=$(curl -fs https://royaltsx-v4.royalapps.com/updates_stable | xpath '//rss/channel/item[1]/enclosure/@url'  2>/dev/null | cut -d '"' -f 2)
         expectedTeamID="VXP8K9EDP6"
         ;;
-        
+    docker)
```
kenchan0130, line 470: @scriptingosx 
Docker.app need install additional components with admin privilege in the app.
Is that outside the scope of this script?

scriptingosx: yes, most definitely. I recommend building a custom pkg or script for this app that installs the required components _after_ Installomator did the base install. There are many apps that require this extra step, like `bbedit` or `privileges` but the steps to install the additional components may be different for each

securitygeneration: The process of getting Docker onto the machine is just a DMG-app-drag-drop. I just want to get the binary onto the endpoint. The app itself takes care of the rest, and most of my users will have the permissions to run it.

kenchan0130: i understood. Thank you for the explanation.

PR #22, opened 2020-05-23
Introduce CI

kenchan0130:
It's hard to manually verify that each application can be installed.
Therefore, we have introduced CI to automate this process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,20 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "!*"
+  pull_request:
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run test
+        run: |
+          concurrency=$(sysctl -n hw.ncpu)
+          cat Labels.txt | parallel --halt now,fail=1 --jobs ${concurrency} --tagstring {}: ./Installomator.sh {}
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run test
+        run: |
+          concurrency=$(sysctl -n hw.ncpu)
+          cat Labels.txt | parallel --halt now,fail=1 --jobs ${concurrency} --tagstring {}: ./Installomator.sh {}
+        env:
+          GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
kenchan0130, line 20: Once the solution of [this problem](https://github.com/scriptingosx/Installomator/pull/4#issuecomment-633017326) is done, it will work effectively.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,20 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
```
kenchan0130, line 5: As far as the master branch is concerned, it might be better to just run it periodically instead of using push events.

kenchan0130:
I didn't get much interest from you guys, so I'll close.

Theile:
We are not that active, so we are not ignoring it, or disliking it, we just didn't come to this thing now.

Can you explain what it is you suggested. I don't know GitHub enough to know for sure what change you suggest.

Thank you!

PR #28, opened 2020-06-11
Install sketch

loceee:
added Sketch

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -8,8 +8,8 @@
 # inspired by the download scripts from William Smith and Sander Schram
 # with additional ideas and contribution from Isaac Ordonez, Mann consulting
 
-VERSION='0.2'
-VERSIONDATE='20200529'
+VERSION='0.3'
+VERSIONDATE='20200609'
 
 export PATH=/usr/bin:/bin:/usr/sbin:/sbin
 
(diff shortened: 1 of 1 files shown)


Installomator.sh
```diff
@@ -679,7 +679,15 @@ case $label in
         downloadURL="https://updates.cdn-apple.com/2019/cert/061-41823-20191025-5efc5a59-d7dc-46d3-9096-396bb8cb4a73/SwiftRuntimeForCommandLineTools.dmg"
         expectedTeamID="Software Update"
         ;;
-
+    sketch)
+        name="Sketch"
+        type="zip"
+        downloadURL=$(curl -sf "https://www.sketch.com/get/" \
```
kenchan0130, line 685: Html parsing is subject to frequent changes, so we should avoid it if possible.
It seems better to use https://download.sketchapp.com/sketch-versions.xml instead of it.

loceee: For sure. Totally missed the sparkle feed. Will update and make a new PR.

loceee: ...and of course the Sketch feed is in reverse order comparing it the other sparkle feeds. It would be good to implement that generic sparkle feed function. I might take a pass at it.

kenchan0130: It seems to be listed as an issue in https://github.com/scriptingosx/Installomator/issues/19.

Issue #29, opened 2020-06-12
Adobe Reader DC download fails + installer complications


fponcelin:
I've noticed three problems with the Adobe Reader DC item:

1. The download URL appears to be incorrect

```
2020-06-12 13:17:17 ################## adobereaderdc
2020-06-12 13:17:17 Changing directory to /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.8tXeI4TN
2020-06-12 13:17:17 found app at /Applications/Adobe Acrobat Reader DC.app, version 19.008.20071
2020-06-12 13:17:17 no more blocking processes, continue with update
2020-06-12 13:17:17 Downloading http://ardownload.adobe.com/pub/adobe/reader/mac/AcrobatDC/2000920067/AcroRdrDC_2000920067_MUI.dmg to Adobe Acrobat Reader DC.dmg
2020-06-12 13:17:18 error downloading http://ardownload.adobe.com/pub/adobe/reader/mac/AcrobatDC/2000920067/AcroRdrDC_2000920067_MUI.dmg
2020-06-12 13:17:18 Deleting /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.8tXeI4TN
2020-06-12 13:17:18 ################## End Installomator
```

I found this URL instead (from [Adobe Enterprise Toolkit](https://www.adobe.com/devnet-docs/acrobatetk/tools/ReleaseNotesDC/continuous/dccontinuousmay2020qfe2.html#dccontinuousmaytwentytwentyqfe2)):
`https://ardownload2.adobe.com/pub/adobe/reader/mac/AcrobatDC/2000920067/AcroRdrDCUpd2000920067_MUI.dmg`

2. The package from the URL I found can only update an existing install on the system

From the Installomator log:
```
2020-06-12 14:58:02 ################## adobereaderdc
2020-06-12 14:58:02 Changing directory to /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.UW7JKKZI
2020-06-12 14:58:02 could not find Adobe Acrobat Reader DC.app
2020-06-12 14:58:02 no more blocking processes, continue with update
2020-06-12 14:58:02 Downloading http://ardownload2.adobe.com/pub/adobe/reader/mac/AcrobatDC/2000920067/AcroRdrDCUpd2000920067_MUI.dmg to Adobe Acrobat Reader DC.dmg
[...]

PR #30, opened 2020-07-02
To resolve issue #29 AdobeReaderDC Failing

noelmom:
Changed the way to obtain the latest version from "https://armmf.adobe.com/arm-manifests/mac/AcrobatDC/reader/current_version.txt" since the current version is not always a "full" version.  Instead, I'm using https://get.adobe.com/reader always displays the latest full version.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -594,9 +594,10 @@ case $label in
         expectedTeamID="PS2F6S478M"
         ;;
     adobereaderdc)
+    	# credit: Noelmo Melo
         name="Adobe Acrobat Reader DC"
         type="pkgInDmg"
-        downloadURL=$(adobecurrent=`curl -s https://armmf.adobe.com/arm-manifests/mac/AcrobatDC/reader/current_version.txt | tr -d '.'` && echo http://ardownload.adobe.com/pub/adobe/reader/mac/AcrobatDC/"$adobecurrent"/AcroRdrDC_"$adobecurrent"_MUI.dmg)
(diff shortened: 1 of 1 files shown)


fponcelin:
This works great to get the full installer, thanks! However if using Installomator for patch management, then the MDM will keep installing it over and over since that full installer is not necessarily the latest published version.
I modified and tested the code block with this and it seems to do the trick for both fresh install and patch management:
```
    adobereaderdc)
        name="Adobe Acrobat Reader DC"
        type="pkgInDmg"
        if [[ -d "/Applications/Adobe Acrobat Reader DC.app" ]]; then
        	downloadURL=$(adobecurrent=`curl -s https://armmf.adobe.com/arm-manifests/mac/AcrobatDC/reader/current_version.txt | tr -d '.'` && echo https://ardownload2.adobe.com/pub/adobe/reader/mac/AcrobatDC/"$adobecurrent"/AcroRdrDCUpd"$adobecurrent"_MUI.dmg)
        else
        	downloadURL=$(adobecurrent=`curl -s -H "User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15)" https://get.adobe.com/reader/ | grep "<strong>Version" | sed 's/[^0-9]*//g' | cut -c 3-` && echo http://ardownload.adobe.com/pub/adobe/reader/mac/AcrobatDC/"$adobecurrent"/AcroRdrDC_"$adobecurrent"_MUI.dmg)
        fi
        expectedTeamID="JQ525L2MZD"
        blockingProcesses=( "AdobeReader" )
        ;;
```

scriptingosx:
I believe this has been resolved in v0.3

PR #31, opened 2020-07-02
Support aquaskk

kenchan0130:
The aquaskk is IME for mac like Google Japanese Input.

https://github.com/codefirst/aquaskk


scriptingosx:
added to dev branch in 5a60517

PR #32, opened 2020-07-02
Support Krisp

kenchan0130:
The Krisp is a tool for advanced noise cancellation.

https://krisp.ai/


scriptingosx:
added in b4c9873

PR #36, opened 2020-07-04
Version checking before installation

Theile:
If a label contains the variable "appNewVersion", then this version will compare this version to the installed version, and only demand an installation if there is no match.


Theile:
I cannot take credit for adobereaderdc, I only found the version number.

PR #39, opened 2020-07-04
If Spotlight is not returning the app

Theile:
Then we can search manually in /Applications.


Theile:
I have investigated this issue. It looks like when the Mac is set to Danish, then kMDItemKind will not return "Application" but the Danish translation "Program". I wander if "kind:application" can be used instead. 

(At first I thought it was my machine with a problematic Spotlight database, but i get this error from other machines as well. Strange that it has been translated at all).

PR #50, opened 2020-07-09
Added an extra parameter to script

Theile:
I would like to override BLOCKING_PROCESS_ACTION=prompt_user, so an extra parameter will do that.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -183,6 +183,29 @@ printlog "################## $label"
 # lowercase the label
 label=${label:l}
 
+# Second parameter as the handling of blocked process (so admin can force update)
+# credit: Søren Theilgaard (@theilgaard)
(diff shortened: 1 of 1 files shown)


Raptor399:
I'd say apply the same logic to all three major configuration settings: `DEBUG`, `NOTIFY` as well as `BLOCKING_PROCCESS_ACTION`. You'd want to configure these as script parameters in Jamf without having to change the source code.

scriptingosx:
I have started working on a more generalized version of this in the `argument-refactor` branch.

The changes i did so far now allow to provide and override variables in the form `VAR=value` as an argument:

```
% ./Installomator.sh NOTIFY=silent BLOCKING_PROCESS-ACTION=ignore desktoppr
```

all the variables as arguments have to be provided _before_ the label

PR #54, opened 2020-07-24
Need echo for the Yes to DMG mounts

Theile:
A little bug, or one copy-paste too many of printlog…:-)


scriptingosx:
Thank you, fixed in 0a96de8

PR #55, opened 2020-08-04
added parallels desktop

aduffner:
Added the label "parallels" for Parallels Desktop (business edition) in version 15 (downloads the latest of version 15) - could add some logic to find out if 16 exists or greater

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -203,6 +203,16 @@ longversion)
     ;;
 
 # label descriptions start here
+
+# TODO: parallels installation process needs testing
+# parallels)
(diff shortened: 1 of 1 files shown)


scriptingosx:
This is an installer app rather than the final application itself. So far, Installomator does not support this kind of installation.

There have been requests to implement this kind of installation, though. Will keep this until we start working on that.

Theile:
This actually downloads an installer app, and Installomator cannot handle that. I have made a label file in our "disabled" folder, then we have saved the work that has already been made.

PR #58, opened 2020-08-23
Only match GitHub files that end in the filetype

Raptor399:
Some GitHub repos (e.g. keepassxreboot/keepassxc) `downloadURLFromGit()` returns multiple files matching the filetype, e.g. `filename.dmg`, `filename.dmg.DIGEST`, etc. and installation will fail.

By requiring the end quote to match, the function only returns the one file that ends in the filetype and installation will succeed.


scriptingosx:
thanks !

PR #77, opened 2020-10-11
Prompt the user only once

Raptor399:
Since the prompt is in a for loop, it could be triggered 3 times if the process
does not exit in time. This change makes the script only prompt once, then
remembers the button.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -249,7 +249,7 @@ checkRunningProcesses() {
                       printlog "killing process $x"
                       pkill $x
                       ;;
-                    prompt_user)
+                    prompt_user|prompt_user_then_kill)
                       # prompt the user once
                       if [[ $button = "" ]]; then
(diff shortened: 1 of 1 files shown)


scriptingosx:
not entirely convinced of this (yet). The reason I _wanted_ to show the dialog multiple times, is that if the process has not quit after 30 seconds, it will likely require some user attention to fix. (i.e. accept file save dialogs)

Theile:
We are now working on dev 0.5. It's very much different from version 0.4, and now this pull request is in conflict.
Could you please test the new dev and see if you still think you need this (there are also changes to this). Then I would ask you to create a new pull request.
Thank you!

Raptor399:
Quickly scanning the code of 0.5, the user will still be prompted multiple times when clicking "Quit" and the application fails to quit. But you're right in closing the pull request, because the solution is no longer valid for 0.5.

Theile:
You are very welcome to suggest a replacement…

PR #78, opened 2020-10-11
Add option to prompt user and finally kill

Raptor399:
The `prompt_user` option uses osascript to attempt to quit an app three times,
which can sometimes fail. Added an option `prompt_user_then_kill` to attempt
to quit twice, but kill the process if earlier quit attempts failed.


scriptingosx:
thank you!

PR #95, opened 2020-11-23
Added label for Android File Transfer

saess-sep:
I added Android File Transfer, after testing it with the application closed and open.  The installation and blocking process work.


scriptingosx:
Thanks!

PR #97, opened 2020-11-24
Added option to prompt, then quit directly

saess-sep:
MS Teams specifically was not responding to the quit option, and prompting three times seemed a bit much, so I added this option.  There's another PR (#77) to change the prompt_user_then_kill functionality, but I think giving the option is the way to go.


Theile:
We are working on a new dev 0.5. That unfortunately makes this pull request in conflict.

I can see the idea of this request, but please take a look at the new dev, as there is other handlings. I don't believe Teams to have a general problem with not willing to quit, so I have not encountered the need for this.

Please reevaluate at let us know what you think. Thank you!

saess-sep:
I'll try out the dev branch and see if MS Teams and Dropbox still don't respond kindly to being asked to quit.  Either way, this can be closed if it's not the direction the script is heading in.

Theile:
Thanks!

PR #99, opened 2020-12-04
Brave Browser: Fixed xpath

midni9ht:
Changed xpath from "item[1]" to "item[last()]", so the latest version of Brave is downloaded.


scriptingosx:
Thank you!

PR #113, opened 2021-01-11
Fixed DBeaver URL

midni9ht:
DBeaver now ships as DMG (was a PKG before). Fixed the URL for that matter, so download works again.


midni9ht:
This has been implemented by another pull request: https://github.com/scriptingosx/Installomator/pull/131

PR #125, opened 2021-01-22
Slack for Apple Silicon, too

Theile:


arnoldtaw:
Even Slack universal was added on v0.4.5, I'm on v10.2 and my Silicon mac keeps downloading Intel version. 

v10.3 changes does not include anything for Slack.

PR #126, opened 2021-01-22
Adding Rocket.Chat, FontExplorer X Pro and nextcloud

Apfelpom:


Theile:
rocketchat and nextcloud has been added!
FontExplorer is not working out. It needs a filling of form, and I cannot see we can get the link out. If I fill it out and start the download, I get this URL: http://fast.fontexplorerx.com/FontExplorerXPro726.dmg
That is not bad, but it will not be a link to the latest version. Then I simply suggest to call Installomator with this:
`valuesfromarguments name="FontExplorer\ X\ Pro" type="dmg" downloadURL="http://fast.fontexplorerx.com/FontExplorerXPro726.dmg" appNewVersion="7.2.6" expectedTeamID="2V7G2B7WG4"`
But thank you for the other 2 labels!

PR #127, opened 2021-02-04
Added 'element'

midni9ht:
Added the Element Messenger app to Installomator and Labels.


midni9ht:
Element has been implemented in the meantime.

PR #128, opened 2021-02-10
Update Google Drive File Stream pkg Name

kris-anderson:
Google Drive File Stream changed the PGK file name within the DMG

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -855,7 +855,7 @@ googledrivefilestream)
     name="Google Drive File Stream"
     type="pkgInDmg"
     downloadURL="https://dl.google.com/drive-file-stream/GoogleDriveFileStream.dmg"
-    pkgName="GoogleDriveFileStream.pkg"
+    pkgName="GoogleDrive.pkg"
     expectedTeamID="EQHXZ8M8AV"
     ;;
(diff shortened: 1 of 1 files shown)


apizz:
It would also be helpful if this change included options to use either `googledrivefilestream` (to avoid breaking existing configs) and `googledrive` to trigger this item.

Theile:
This label has been improved in 0.5 dev. I don't think we need to name the pkg-file. So currently it will work what we have added already, and I will proceed to close this. Thank you!

quantumadvisors:
I found this issue and saw that the pkg name was kept the same (both URLs seem to work), but the change of the app name from "Google Drive File Stream" to "Google Drive" wasn't implemented. It does look like the app is called 'Google Drive.app' now so the script is not finding it to check version, re-open after installation, etc.

sswany:
I found this issue and saw that the pkg name was kept the same (both URLs seem to work), but the change of the app name from "Google Drive File Stream" to "Google Drive" wasn't implemented. It does look like the app is called 'Google Drive.app' now so the script is not finding it to check version, re-open after installation, etc. 

Is that something that could be changed in a future release? Thanks!

PR #130, opened 2021-02-15
With Theile additions (0.5.0)

Theile:


Theile:
With LOGO variable, and removal of re-launch as it gave problems with keychain.

PR #131, opened 2021-02-15
Fix broken DBeaver install

darklink87:
PKG is no longer available and the download link is no longer functional. Updating it to use DMG with the updated download link.


isaacatmann (approved):
Thank you!

PR #132, opened 2021-02-22
Added IntelliJ IDEA + Fixed Google Drive File Stream

darklink87:
Added the paid version of IntelliJ IDEA, otherwise known as IntelliJ IDEA Ultimate on https://www.jetbrains.com/idea/download/#section=windows


Theile:
I believe we have fixed googledrive in our latest dev 0.5. You are very welcome to test that.
The other label has been added, but with name "jetbrainsintellijidea".
Thank you very much!

PR #135, opened 2021-03-17
Fixed download for Docker Desktop

midni9ht:
The download URL for the latest version of Docker Desktop has changed to https://desktop.docker.com/mac/stable/amd64/Docker.dmg


Theile:
To me it looks like both URLs downloads the same stuff. I just downloaded them, and they are identical and offers an Intel only app. I will proceed to close this, thank you!

PR #138, opened 2021-03-19
Added 'viscosity'

midni9ht:
Added Viscosity VPN to Installomator and Labels.


midni9ht:
Visosity was implemented with the last update of Installomator (by @matins).

PR #140, opened 2021-03-31
Add timeout to prevent blocking other processes

Raptor399:
Added variable `TIMEOUT` that can be set to the number of seconds.

When the timeout expires, the dialogue box will give up waiting for a button click.
This prevents the script from blocking other processes while the user ignores the
dialogue box for a prolonged period of time.

Although TIMEOUT should probably not be empty by default, I kept it that way to
not introduce changes to the existing script and stick to its earlier behavior.

Inspired by @saess-sep commit https://github.com/scriptingosx/Installomator/pull/97/commits/d6c255d2c6247c9a947749dba3af0f72bf4f7b20


scriptingosx:
overlaps with #771

Raptor399:
Correct. #771 is the light version of this PR, quietly managing the "Not now" variant. This PR messages the user what's about to happen and differentiates its behaviour based on the preferred type of action.

Either PR will remedy the problem of the Installomator script blocking its parent process due to user inaction, which makes the execution of Jamf policies unreliable. It would be really nice if one of the PRs could be included in a future release.

PR #151, opened 2021-04-15
Fix Plantronics Hub & add Hanock & iMazing profile editor

pro4tlzz:
Change type to fix  Plantronics Hub


pro4tlzz:
Also add support for iMazing Profile Editor

Theile:
Thanks, looks good.

PR #152, opened 2021-04-15
Add Arq 7

micahflee:
Arq 7 is backup software: https://arqbackup.com/download/

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -2993,6 +2993,14 @@ microsoftyammer)
     #updateTool="/Library/Application Support/Microsoft/MAU2.0/Microsoft AutoUpdate.app/Contents/MacOS/msupdate"
     #updateToolArguments=( --install --apps ?????? )
     ;;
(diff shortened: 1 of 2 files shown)


Theile:
I have included this in the dev-version. Thank you!

Next time when you have the URL, try using the buildCaseStatement script, as it will get the packageID, which is a very good parameter to check for version number, if it has already been installed. And if you find a web page with the version in, let us know about that, so we can built the appNewVersion variable.

```
arq7)
    name="Arq7"
    type="pkg"
    packageID="com.haystacksoftware.Arq"
    downloadURL="https://arqbackup.com/download/arqbackup/Arq7.pkg"
    appNewVersion="$(curl -fs "https://arqbackup.com" | grep -io "version .*[0-9.]*.* for macOS" | cut -d ">" -f2 | cut -d "<" -f1)"
    expectedTeamID="48ZCSDVL96"
    ;;

```

Issue #156, opened 2021-04-26
Webex teams failing


Simply-Guy:
Installing Webex Teams seems not to work. I think the .dmg file is no longer available without logging in.

midni9ht:
Hey @Simply-Guy 
Webex Teams has been renamed to just 'Webex.app', which caused Installomator to fail, because the application name could not be fount within the downloaded DMG.
I just created a pull request (#164 ) to reflect that change and also add support for Apple Silicon at the same time.
This should solve the problem for you ;-)

Theile:
Thanks, I think that has been fixed now.

PR #164, opened 2021-05-07
Updated 'webexteams' for Apple Silicon support

midni9ht:
'Webex Teams' changed its application name to just 'Webex'.
Also there is an Apple Silicon version available now.
Should solve #156

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -2582,9 +2582,13 @@ webexmeetings)
     ;;
 webexteams)
     # credit: Erik Stam (@erikstam)
-    name="Webex Teams"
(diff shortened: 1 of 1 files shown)


Theile:
Thanks!

So it is not Webex Teams anymore, only Webex? Is it the same app, is it just renamed, or do you know what happened?

midni9ht:
Hey @Theile 
Yes, it looks like the app has been renamed. There's no reference to "Webex Teams" on their website anymore. The app itself appears to be basically the same.

Theile:
All right, thanks!

PR #166, opened 2021-05-11
vmware

mathieu244:


Theile:
Can you explain some of these changes, please.
What labels need the rename app?
and this line:
`if [[ $appversion == *"${appNewVersion}"* ]]; then`
What do you expect to get by this change?

In general we would like label changes to be separate from other changes, if you can separate those, it would be great.

Thanks!

Theile:
Seems like this PR has grown a lot. I don't think it will be vice for us to add a script-running mechanism to this project. And if we did something like this, we would not name it anything with bash, as we use zsh, so more like "shell" (after all it's also a .sh file it is looking for). But the scripts it is executing does not have any signatures, so it would be introduction of a feature that would circumvent previous effort.

The label haxm is not even downloading haxm, that is only part of the comments in the label. And haxm is actually a pkgInDmgInZip, and we do not have a mechanism for that, but not sure we need to run any script from the zip-file to run that installer anyway.

And why was all this put in a VMware PR. 

I would think that all of this looks as so specialised features, it should be handled manually, but I will ask the others…

Theile:
Did you read my last comment?
This PR has too many changes, and we are not prepared to integrate all of this.
Please make separate PRs for new labels, and separate PRs for feature changes. This is too much in one PR.
I have copied the insomnia label out for a future release.
I will close this PR.

PR #169, opened 2021-05-13
Fix to install googlejapaneseinput

kenchan0130:
Google Japanese Input app will be restarted by the launch daemon when the application is terminated.

The `checkBlockingProcesses` process kills the corresponding process and sleeps for a few seconds, so the behavior of the launch daemon causes an infinite loop.

Therefore, I thought it would be useful to avoid blockingProcesses as `NONE` for this application.


Theile:
Thank you, I trust this input.

PR #171, opened 2021-05-14
dont use msupdate for microsoftteams

sphen13:
msupdate does not fully support teams


Theile:
Thank you, I have also seen som weird behavior.

PR #172, opened 2021-05-17
Dont download pkg if we are set to silent_fail and there is blockingprocess

sphen13:
Fixes issue where installomator will repeatedly download the pkg each run when set to silent_fail and a blockingprocess is present.


scriptingosx:
Thank you!

PR #187, opened 2021-06-18
Zulu JDK fixes

kryptonit1:
Zulu JDK 11, 13, 15 fixed and Zulu JDK 8 added.


acodega:
Could you try redoing this pull, you have many line changes being submitted which shouldn't be, and the PatchSkip and Company variables (whether or not they're commented out) are no longer applicable.

PR #195, opened 2021-06-22
Added 'appNewVersion' to 'element'

adibue:
Added version checking via GitHub to `element`.


acodega:
I'm a little nervous about checking the version off Git but downloading it from the vendor's website, but as long as they maintain parity in the future it's fine of course.

PR #197, opened 2021-06-23
Got 'mattermost' ready for Apple Silicon

adibue:
`mattermost` can now distinguish between Intel and Apple Silicon.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -1938,7 +1938,13 @@ malwarebytes)
 mattermost)
     name="Mattermost"
     type="dmg"
-    downloadURL=$(downloadURLFromGit mattermost desktop)
+    if [[ $(arch) == i386 ]]; then
+      downloadURL=$(curl --silent --fail "https://api.github.com/repos/mattermost/desktop/releases/latest" \
+      | awk -F '"' "/browser_download_url/ && /mac.dmg/ && ! /blockmap/ { print \$4 }")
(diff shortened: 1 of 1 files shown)


acodega (changes requested):
I tested this and it works for instances where a GitHub repo has different Release URLs for different platforms. Going with this would allow `downloadURLFromGit` to still do it's thing. So less code on your part. ```mattermost)
    name="Mattermost"
    type="dmg"
    if [[ $(arch) == i386 ]]; then
      archiveName="mac-m1.dmg"
    elif [[ $(arch) == arm64 ]]; then
      archiveName="mac-m1.dmg"
    fi
    downloadURL=$(downloadURLFromGit mattermost desktop)
    appNewVersion=$(versionFromGit mattermost desktop )
    expectedTeamID="UQ8HT4Q2XM"
    ;;```

adibue:
Hey @acodega 
That's great! I just saw today, that there's the option `archiveName`. Is that new or did I just oversee it?
Anyways, thanks for the input :-)

I gues with
`    if [[ $(arch) == i386 ]]; then
      archiveName="mac-m1.dmg"`
the `archiveName` should be "mac.dmg"?

acodega:
I think archiveName has been around for some time, from the readme `When not given the archiveName is set to $name.$type` I'm not sure the original intention but it works perfectly for separate Intel and M1 builds on GitHub.

PR #198, opened 2021-06-24
Support variables from script caller

kenchan0130:
According to the readme, you define a global variable like 

```sh
./Installomator.sh firefox LOGO=jamf BLOCKING_PROCESS_ACTION=tell_user_then_kill NOTIFY=all
```

and call the script.

However, in practice, this does not work as expected.

This is because the variables are overwritten inside the script.

Assuming the readme is correct, I made it so that we can set some global variables when calling the script.

And, 

```sh
./Installomator.sh firefox LOGO=jamf BLOCKING_PROCESS_ACTION=tell_user_then_kill NOTIFY=all
```

 is not recognized correctly as a shell variable.
Therefore, I changed this project readme description to recognize variables correctly like 

```sh
LOGO=jamf BLOCKING_PROCESS_ACTION=tell_user_then_kill NOTIFY=all ./Installomator.sh firefox
```

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -20,10 +20,10 @@ export PATH=/usr/bin:/bin:/usr/sbin:/sbin
 # set to 0 for production, 1 for debugging
 # while debugging, items will be downloaded to the parent directory of this script
 # also no actual installation will be performed
-DEBUG=1
+DEBUG="${DEBUG:-1}"
 
 # notify behavior
-NOTIFY=success
+NOTIFY="${NOTIFY:-success}"
 # options:
 #   - success      notify the user on success
 #   - silent       no notifications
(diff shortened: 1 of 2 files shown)


scriptingosx:
You are misunderstanding how this is supposed to work. The intention of this feature is *not* to read environment variables.

Some management systems allow the admin to pass a number of arguments when calling a script. Since an admin may want to override any number of arbitrary variables used in the script, even variables that we haven't implemented yet, positional arguments won't work.

None of the management systems I know, allow to set environment variables for the context of script (though I admit that would be neat).

Instead I put in [some code](https://github.com/Installomator/Installomator/blob/01858272565e5b9ca45b6b4887b27745bda0f7b6/Installomator.sh#L802-L813) which parse the arguments and when an argument contains an `=` it passes it through `eval` which sets the variable. Since this happens after everything is initialized, it will override all the presets from earlier in the code.

kenchan0130:
Thank you for the explanation.
I now understand the behavior of this script.

PR #200, opened 2021-06-27
Look in /Application and Utilities first

Theile:
First /Applications, then /Applications/Utilities, and then Spotlight. Maybe we should be prepared that DEPnotify should be installed in Utilities, so `customDestination="/Applications/Utilities/"`could ba a thing.


scriptingosx:
Thanks!

PR #192, opened 2021-06-21
Fixed 'webexmeetings' URL

adibue:
The URL for `webexmeetings` appears to have changed yet another time.
Good thing: The downloaded `pkg` contains the versions for Intel and Apple Silicon at the same time. So no additional URL required :-)

PR #206, opened 2021-07-12
Added 'webex' to 'webexteams'

adibue:
Replaces #192.
Added `webex` as Label to `webexteams`, since those are identical.

NOTE: As of a reply I got from Cisco support, Cisco is consolidating "Cisco Webex Meetings" and "Cisco Webex Teams" into just "Webex".
The download of the Apple Silicon version of Webex Meetings does not work reliably (fails most of the time). So I will no longer consider this one.


scriptingosx:
Thanks!

PR #221, opened 2021-07-18
keyboardmaestro doublet removed

Theile:
Kept the original version (that I actually created with `appNewVersion`).


scriptingosx:
fixed in the fragments branch

PR #228, opened 2021-07-26
Adding Box Tools

Raptor399:
From: https://support.box.com/hc/en-us/articles/360043695834-Large-Scale-Deployments-Box-Tools


abacuswealth:
I've been trying to use this addition but noticed that it was reinstalling each time because it couldn't find the 'Box Tools' app (since its buried and called 'Box Edit'). I got around it by adding the following to our command:

`packageID="com.box.boxtools.installer.boxedit"` 

Thought it may be helpful for anyone else.

PR #249, opened 2021-09-08
BLOCKING_PROCESS_ACTION=tell_user

Theile:
Default now is `BLOCKING_PROCESS_ACTION=tell_user`.

I think this is the best alternative.


scriptingosx:
Thank you

PR #252, opened 2021-09-10
jetbrains apps - make all arch aware downloads

virtualpeter:


Theile:
Can you remove the double-name of the label? We should only have one. The ones that currently has a double-name is because it was renamed, and all JetBrains apps is now called something that begins with JetBrains, just like those from Microsoft.

And could you move this to our new fragments system, please. And you are welcome to update the other JetBrains apps as well.

Thank you!

Theile:
And BTW, we are in the process to remove credit lines. We try to reduce size…

virtualpeter:
oh I see - I shouldn't push changes to the assembled script?   I will do a new pull request then, I have a different idea now for the code.

PR #254, opened 2021-09-17
Added Keybase

fleish:


Theile:
Thanks for this! Unfortunately you have changed the wrong file. instead of changing Installomator.sh, you should add a new file in `fragments/labels`, like `fragments/labels/keybase.sh`.

I took a look at the label, and I think for this label, I would either use this for downloadURL …
`curl -s https://keybase.io/docs/the_app/install_macos | grep data-target | cut -d '"' -f2`

… or simply the URL directly, as that is probably being redirected to the latest version (somehow):
`https://prerelease.keybase.io/Keybase.dmg`

But please put it in a file, and then we can merge it, thank you!

fleish:
I actually added the file `fragments/labels/keybase.sh` in this commit as well as updated `Labels.txt` in addition to updating `Installomator.sh`. I did this based on some previous pull requests that appeared to update all 3 as well so I thought that was the right way. I can update as needed but can you clarify if I also need to back-out the entry in `Labels.txt`?

It might be helpful to have a quick note about the process if one does not exist currently.

Theile:
`Labels.txt`is fine to update. So then just fix `downloadURL`, and remove the changes to `Installomator.sh`.

Thanks!

fleish:
OK, should be done

Theile:
Thank you very much!

Theile:
BTW, we try to keep the script short so we are removing the credit lines. It's not that we are not grateful, but we try to save lines. All labels we edit from now on, will have these lines removed.

PR #256, opened 2021-09-18
Fixes and New Labels

kdrwygvh:
Fix for Jabra Direct and new Labels for Jamf Connect Configuration, Soundsource, Tag Editor, and M4A Converter


Theile:
Thank you for this, but can you move your additions to the fragments parts. The new system is that we need the fragments updated as those are the basis for the build of the full script. 

You just create new labels files as needed.

Thanks!

PR #257, opened 2021-09-20
Added Sync

fleish:


Theile:
Thank you, I added `appNewVersion` and removed the credit-line (we don't include these anymore). Thanks!

PR #268, opened 2021-10-12
Added option to not interrupt full screen apps like KeyNote or Zoom

Raptor399:
We've all had that call from one of our colleagues, pleading to stop interrupting their presentation with our bothersome scripts. Configure `INTERRUPT_DND="no"` and the script will try to detect apps with an active display sleep assertion. If one is found, the script will abort with an error, instead of embarrassing the user in front of their audience.

A list of apps whose display sleep assertions can safely be ignored `IGNORE_DND_APPS` has been added as well, e.g., so playing a YouTube video won't block the script from running.


Theile:
This looks very interesting. 

What does DND mean, just to understand that?

We have also talked about detecting use of apps if they used more than 5% of CPU, but a full screen detection would also be great.

Theile:
I see, Do Not Disturb!

But maybe IINTERUPT_FULLSCREEN is better then…

if app has appNewVersion defined I actually think it would be great to know if an updated was needed at all, as we then can exit with success. 

I will look at this later.

smilieK:
Yeah in my testing no matter what you're doing in a Web Browser as long as it's in the front browser window it will show the same assertion command if in video/Audio/Web based meeting platform. So it's a hard way to tell if user in a web based meeting or just watching music on YouTube. :(

pid 46672(Safari): [0x0002bde100058888] 00:02:04 PreventUserIdleDisplaySleep named: "com.apple.WebCore: HTMLMediaElement playback"
pid 92525(firefox): [0x0002d21200058c78] 00:05:05 NoDisplaySleepAssertion named: "video-playing"
pid 92813(Google Chrome): [0x0002d38300058cda] 00:05:02 NoDisplaySleepAssertion named: "Video Wake Lock"
pid 92892(Microsoft Edge): [0x0002d3c700058d37] 00:03:01 NoDisplaySleepAssertion named: "Video Wake Lock"

kdrwygvh:
Applications like Adobe Premiere and Final Cut are also good candidates for this though I don't know if they create sleep assertions. They are definitely applications where breaking the concentration of the user with a notification should be discouraged.

Theile:
I like this idea and have no further questions…

brunerd:
Sorry to revive a dead thread but I was wondering about the `pmset -g | awk...` code ([link](https://github.com/Installomator/Installomator/blob/a3dbf75818bec8d2f1fdc91d490ec5cf67745308/fragments/functions.sh#L834))? 

It's really clever and I wanted to use it with proper attribution to this project and @Raptor399 and then just today I saw the same code used [here](https://github.com/Macjutsu/super/blob/2634d20ab58fb7f5698800fe343ef8ffa77afbcf/super#L1940) just today!

I'm guessing _this one_ is the originator since the commit date is way further back but are there any other misty origins to this one-liner? Or have I found the source? If so: Thanks! I'd like to add it to my page [Respecting Focus and Meeting Status in Your Mac scripts (aka Don’t Be a Jerk)](https://www.brunerd.com/blog/2022/03/07/respecting-focus-and-meeting-status-in-your-mac-scripts-aka-dont-be-a-jerk/) with attribution of course :)

Theile:
This line is from @isaacatmann. Not sure where he got it from.

Raptor399:
That brilliant line is definitely not mine, don't credit me! I picked it up from some other script, but I can't recall its origins or author, sorry. Pretty sure it wasn't the code @brunerd linked.

Issue #88, opened 2020-11-18
check for root earlier when not in debug mode


SmithersJr:
When testing a new label, I test locally in debug mode first. If everything looks good there, I test locally with debug mode off before loading it up in Jamf. If I forget to run as root, I have to wait for the whole download, verification, etc. before the error, `ERROR: not running as root, exiting`. 

It would be helpful to get that error at the beginning of the process, so I don't have to wait another 8 minutes to test. :-/

(And, also, because I'm a goober that can't remember to run it with `sudo`.) :-P

Theile:
I love that the root check is so late, as I do a lot of testing with downloading, when not in DEBUG mode, just by running it without root access.

acodega:
I agree with @SmithersJr here, the root check should be one of the first things the script does.

samess-flowers:
Would moving the check forward in the process while adding an additional debug state that does everything but install things resolve this for everyone?  `DEBUG=0` and `DEBUG=1` would work the way they currently do, but a new state `DEBUG=2` (which doesn't feel properly idiomatic, but would fit the pattern) could either skip the earlier root check (the easier way to implement) or purposefully terminate where the current root check is even if the program is being run as root (which feels "cleaner" to me).

PR #278, opened 2021-10-16
Move root user check forward and add in debug mode 2 to replicate lost functionality

samess-flowers:
This is to resolve [Issue 88](https://github.com/Installomator/Installomator/issues/88).

I've moved the root user check forward as requested in the issue, but also added in a new debug mode to replicate the functionality that @Theile was using for testing that was lost when I move it forward.  Debug mode two skips most of the debug checks, except where the old root user checks were, at which point the script ends just as it would have previously.  I tested this in all three debug modes with a few different labels and it worked on my machine.  Please let me know if there's any other information needed.


Theile:
This looks really great. I Have moved the root check to before the case statement, as we can have code in that (see microsoft labels). So if we were supposed to exit, then we do that before this.
@scriptingosx have you looked at this?

PR #311, opened 2021-11-09
Added packedID so Box Tools installs correctly

### LINKED ISSUE #310: Box Tools not installing correctly
Box Tools is reinstalling each time because it cannot find the 'Box Tools' app (since it's buried and called 'Box Edit'). Get around it by adding the following to the label:

```
packageID="com.box.boxtools.installer.boxedit"
```

_Originally posted by @abacuswealth in https://github.com/Installomator/Installomator/issues/228#issuecomment-961527090_

Raptor399:
Fixes #310

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -1280,6 +1280,7 @@ boxsync)
 boxtools)
     name="Box Tools"
     type="pkg"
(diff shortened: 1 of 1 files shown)


Theile:
Thanks!

PR #331, opened 2021-11-22
changed logic for 'INSTALL=force'

scriptingosx:
The previous logic for `INSTALL=force` would not take when the version from the vendor was newer, than the installed version. Then Installomator would _still_ use the update tool to update. The new logic will download and install when `INSTALL=force` regardless.


Theile:
I don't think we need this. `INSTALL=force` will always force an install, no matter the installed version, and it will make `installerTool=""`.

scriptingosx:
This has been fixed in https://github.com/Installomator/Installomator/blob/b43ebe348324c34cc7ff77e5048c46b117115d8c/fragments/main.sh#L120

PR #346, opened 2021-12-01
possibility to have root:wheel on copied apps

Theile:
On shared machines it can be better to have root:whell as owner of the copied app, rather than using the currently logged in user.

`SYSTEMOWNER=1` will user root:wheel as owner.
`SYSTEMOWNER=0` is default where the current user will be set.

### DIFF
--- a/fragments/header.sh
+++ b/fragments/header.sh
@@ -88,6 +88,13 @@ IGNORE_APP_STORE_APPS=no
 #                  Shouldn’t give any problems for the user in most cases.
 #                  Known bad example: Slack will loose all settings.
 
+# Owner of copied apps
+SYSTEMOWNER=0
+# options:
+#  - 0             Current user will be owner of copied apps, just like if they
+#                  installed it themselves (default).
+#  - 1             root:wheel will be set on the copied app.
+#                  Useful for shared machines.
 
 # install behavior
 INSTALL=""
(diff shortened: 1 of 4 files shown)


fragments/header.sh
```diff
@@ -88,6 +88,13 @@ IGNORE_APP_STORE_APPS=no
 #                  Shouldn’t give any problems for the user in most cases.
 #                  Known bad example: Slack will loose all settings.
 
+# Owner of copied apps
+SYSTEMOWNER=0
```
AustinLeath, line 92: Could `SYSTEMOWNER` be set to `SYSTEMOWNER=1` by default? By setting `SYSTEMOWNER=1` it would make all install scenarios across all computers be equivalent across the board.

AustinLeath: Or even better we could make a flag that lets users set the value of `SYSTEMOWNER` each time they execute the script.

AustinLeath (changes requested):
Here is my take on this PR. You can take a look at the comments to get a better idea of where I am coming from.

fragments/header.sh
AustinLeath:
```suggestion
SYSTEMOWNER=1
```

Theile: No. The history of Installomator is to be like the app was locally installed by the user. You can always sent the variable to the script as a parameter, and it will be set for that run cycle.

AustinLeath: Okay! Thanks for the suggestion.

AustinLeath (approved):
Looks good!

PR #354, opened 2021-12-22
Improve documentation on prompt_user and related variables

acodega:
Some clarity is needed on prompt_user and prompt_user_then_kill, as it sounds like Installomator will continue with the update when the user is prompted. In actuality then_kill refers to how the process will be terminated.


Theile:
I would say that both of them refer to how the app is quit. Either prompt the user, or kill it.

But by all means, I think our documentation can be improved so I love the initiative!

Theile:
`prompt_user` is not default any more `tell_user` is.

Theile:
`tell_user_then_kill`: It will only kill after 3 nice failed attempts.

acodega:
I'll tweak the PR.

Theile:
I think we should get this included in our coming release. Can you fix the conflicts in the merge?

PR #364, opened 2022-01-06
adding prompt with defer option

unfo33:
Suggested change so that you can allow multiple prompts and then force without MDM agent being held up. My intended way of using this is setting it to run once a day with Jamf so the users get prompted a couple times and then are forced.


Theile:
Thank you for the contribution.

Can you somehow remove all the white space changes? Those change all over the file, can make merging of other PRs difficult or impossible.

PR #367, opened 2022-01-07
add prompt with defer and remove whitespaces

unfo33:
Sorry I'm new to contributing. This one looks better it looks like Atom was doing some whitespace things. I wasn't able to see how to overwrite the old one so please just delete the other.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -277,6 +277,43 @@ checkRunningProcesses() {
                           BLOCKING_PROCESS_ACTION=kill
                       fi
                       ;;
+                      prompt_with_defer)
+                        if [ -d /usr/local/Installomator/receipts ]; then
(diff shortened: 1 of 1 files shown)


Theile:
This is super!

I was actually looking more into what this really is about, and I like the solution here.

What I ended up discussing with myself here, is this:
- I don't think the displaydialog should explain to let the user leave the window open, but rather show the countdown on the “Not Now” button.
- If the timer-file is higher than 0, and the app is installed with another `BLOCKING_PROCESS_ACTION`, shouldn't we put the timer to 0 then?
- Maybe plist-files would be better, and then use `defaults write` and `defaults read` for these values. Then all timers can be in one file.
- The `BLOCKING_PROCESS_ACTION` change is to `tell_user_then_kill` but the `displaylog` command only mentions `kill`

PR #368, opened 2022-01-10
Update libreoffice.sh

adibue:
Added `downloadURL` for Apple Silicon. Also added `blockingProcesses`, so open documents are not quit anymore while updating.


Theile:
Cool, thank you!

PR #372, opened 2022-01-12
Isaac logging (awaiting deletion)

Theile:
This PR will implement 4 things:
- Logging levels as DEBUG 0 INFO 1 WARN 2 ERROR 3 REQ 4
- External logging to Datadog
- A function to shorten duplicate lines in installation logs
- Ability to extract install.log in the time when Installomator was running, if further investigations needs to be done to logs


Theile:
This PR is a mess, another will be created.

PR #374, opened 2022-01-12
Fixed 'appNewVersion' for 'docker'

adibue:
URLs for version checking have changed. It's now using the `shortVersionString` from the sparkle feed.

Example run for M1:

```
$ sudo zsh Installomator.sh docker DEBUG=0

2022-01-12 18:16:35 docker setting variable from argument DEBUG=0
2022-01-12 18:16:35 docker ################## Start Installomator v. 8.0
2022-01-12 18:16:35 docker ################## docker
2022-01-12 18:16:35 docker BLOCKING_PROCESS_ACTION=tell_user
2022-01-12 18:16:35 docker NOTIFY=success
2022-01-12 18:16:35 docker LOGO=/System/Applications/App Store.app/Contents/Resources/AppIcon.icns
2022-01-12 18:16:35 docker no blocking processes defined, using Docker as default
2022-01-12 18:16:35 docker Changing directory to /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.t4N46qTN
2022-01-12 18:16:35 docker App(s) found: /Applications/Docker.app
2022-01-12 18:16:35 docker found app at /Applications/Docker.app, version 4.3.2
2022-01-12 18:16:35 docker appversion: 4.3.2
2022-01-12 18:16:35 docker Latest version of Docker is 4.3.2
2022-01-12 18:16:35 docker There is no newer version available.
2022-01-12 18:16:35 docker Deleting /var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmp.t4N46qTN
2022-01-12 18:16:35 docker App not closed, so no reopen.
2022-01-12 18:16:35 docker ################## End Installomator, exit code 0
```

PR #377, opened 2022-01-14
Docker: Changed xpath to use 'last()'

adibue:
When I opened PR #374, the sparkle feed had just one entry. Turns out, that Docker uses the last entry of the feed for the latest version.
With simply changing xpath to use `[last()]`, the version checking now behaves as expected.


Theile:
Thanks for the fix.

PR #386, opened 2022-01-19
Add Code42 arm64 downloadURL introduced with 8.8.2 release

apizz:


apizz:
Looking at this more closely, it unfortunately seems that resulting Apple silicon installer PKG still requires Rosetta to install...

Screen Shot 2022-01-18 at 8 27 48 PM
Screen Shot 2022-01-18 at 8 27 44 PM

Theile:
Thanks, and please remove the credit line. We get rid of those whenever we change a label with those lines, so save a bit of space…

apizz:
done!

Theile:
That was fast, thank you!

PR #389, opened 2022-01-19
Update GPG Suite for blocking process & add DaisyDisk app

fleish:
Looks for GPG Keychain as there is no GPG Suite.app

Add DaisyDisk app from https://daisydiskapp.com/


Theile:
I changed a bit on the `appNewVersion`, but thanks again. This times files are correctly placed.

PR #390, opened 2022-01-19
Added Tableau Public to labels

pythoninthegrass:
> Tableau Public is free software that can allow anyone to connect to a spreadsheet or file and create interactive data visualizations for the web.

Wasn't able to figure out `sed` or `awk` to extract the version from `$downloadURL`; substituted legacy Python2.

Welcome feedback on making more bash/zsh-centric regex and happy to refactor if necessary.

### DIFF
--- a/fragments/labels/tableaupublic.sh
+++ b/fragments/labels/tableaupublic.sh
@@ -3,6 +3,6 @@ tableaupublic)
     type="pkgInDmg"
     packageID="com.tableausoftware.tableaudesktop"
     downloadURL=$(curl -fs https://www.tableau.com/downloads/public/mac | awk '/TableauPublic/' | xmllint --recover --html --xpath "//a/text()" -)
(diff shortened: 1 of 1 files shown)


Theile:
Python is dead. Apple is pretty straight about that. We don't want users to be presented with the dialog of software being used that will not be compatible in the future. So I "fixed" the `appNewVersion`.

Thanks for the label!

pythoninthegrass:
> Python is dead. Apple is pretty straight about that. We don't want users to be presented with the dialog of software being used that will not be compatible in the future. So I "fixed" the `appNewVersion`.
> 
> 
> 
> Thanks for the label!

Oof hahaha. Yeah, they certainly have made life more difficult by not bundling updated runtimes.

Thanks for transpiling the regex in sed. Couldn't get the syntax down, so that's really helpful for future PRs ☺️

Theile:
I had no clue what was going on in that python-part ;-) Learning something new every day, I love it!

pythoninthegrass:
That makes us even hahaha. Appreciate your help @Theile 🙏

PR #408, opened 2022-02-08
Logging level improvements

### LINKED ISSUE #340: FR: Better address scenario in logging where no current installed version of software is detected
Currently when debugging or just reviewing normal run output you get something like the text below when the software that's being installed isn't found to be installed already:

`2021-11-26 01:34:58 textexpander Downloaded version of TextExpander is 7.0.2 (replacing version ).`

The problem is it's not clear if the lack of a version number in `(replacing version )` is because of a bug or legitimately nothing found.  It would be helpful to simply remove this text if we're doing a new install (since other log messages indicate this fact) and only display the `(replacing version X)` message when this is an update of something that already exists.

Theile:
Also fixes #340.


Theile:
In this PR I fix a few mistakes in function `installPkgInDmg`. Somehow we search for `pkgName` in `$tmpDir` and not in `$dmgmount`. Bot only if `pkgName` has been defined in a label.

PR #409, opened 2022-02-08
Trapping the script

### LINKED ISSUE #272: cleanupAndExit with trap
While running the Installomator script, the process may be interrupted for some reason.
For example, the case where the "Self Service" of Jamf policy is canceled by an end-user while it is running.

In that case, the dmg mount and so on may continue to remain, so it would be good if `cleanupAndExit` can be executed even if the process is stopped.

The method doesn't have to use `trap`, but in a shell script, this kind of function has been originally achieved by `trap`.

Theile:
Using these signals:
```
SIGHUP SIGINT SIGQUIT SIGABRT SIGTERM SIGTSTP
```

Added comments in files for which file it is (and which number in the order).
Fixed #272

### DIFF
--- a/fragments/arguments.sh
+++ b/fragments/arguments.sh
@@ -1,4 +1,9 @@
-# MARK: check minimal macOS requirement
+# MARK: 3 Arguments
+
+# Start trapping the script
+trap_with_arg func_trap SIGHUP SIGINT SIGQUIT SIGABRT SIGTERM SIGTSTP
+
+# check minimal macOS requirement
 autoload is-at-least
 
 installedOSversion=$(sw_vers -productVersion)
@@ -125,4 +130,4 @@ valuesfromarguments)
     fi
     ;;
 
-# label descriptions start here
+# Mark: label descriptions start here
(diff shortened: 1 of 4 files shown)


fragments/functions.sh
```diff
@@ -29,6 +29,17 @@ cleanupAndExit() { # $1 = exit code, $2 message, $3 level
     exit "$1"
 }
 
+# Two functions to trap the script
+trap_with_arg() {
+    func="$1" ; shift
+    for sig ; do
```
scriptingosx, line 35: where does the `sig` variable come from?

fragments/arguments.sh
```diff
@@ -1,4 +1,9 @@
-# MARK: check minimal macOS requirement
+# MARK: 3 Arguments
+
+# Start trapping the script
+trap_with_arg func_trap SIGHUP SIGINT SIGQUIT SIGABRT SIGTERM SIGTSTP
```
scriptingosx, line 4: I think it would be easier and clearer to not use the `trap_with_arg` function here, but:

```
for sig in ( SIGHUP SIGINT SIGQUIT SIGABRT SIGTERM SIGTSTP ); do
    trap "func_trap $sig" $sig
done
```

the added abstraction through the extra function is confusing to me

PR #410, opened 2022-02-08
Variable for skipping if app is not installed

### LINKED ISSUE #321: Add an function/ argument for update Apps only?
Hi 

Wish you can consider my suggestion for a new function:

Is it possible to add a function to allow only update the App if it's already installed and do nothing if not installed? This would be useful to run it on all clients to check & update.

I know this can be done by some scripting work, just though it would be nicer to have this function build-in with Installomater.

Thanks a lot!

Luke

Theile:
Fixes #321.


scriptingosx:
could we re-use the `INSTALL` variable for this? if `INSTALL` is `force` it downloads and does a full installation no matter what. When `INSTALL` is `skip_uninstalled` it will abort when the app isn't present yet. When the variable is unset, we get the normal behavior.

acodega:
I think the issue is when: you only want to act if the app is installed **but also** if found you want a force install performed.

As an aside, I think we should choose whether we are going to use yes/no or 1/0 in variable flags and stick to just one format.

scriptingosx:
ok, so we either give the `INSTALL` variable four values: <empty>, `update-only`, `force-update-only` and `force. Or we have two variables, `INSTALL=force` (or empty) and `UPDATE_ONLY=[yes|no|1|0]`

(I do not like `SKIP_UNISTALLED` because it contains _two_ negations ('skip' and 'uninstalled' and is hard to parse semantically)

Mac-Nerd:
Can I ask why use the variables at all, instead of a parameter to switch modes? something like `-i` for installation, `-u` for update only, `-n` for a dry run ("This is what would have happened: these things would have installed, these would have been skipped")

Also, opportunity for a plug for patchomator:
https://github.com/Mac-Nerd/patchomator

PR #427, opened 2022-02-21
Few logging levels changed

Theile:
I want to return a valid explanation for log level REQ for why the script stops.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -444,7 +444,7 @@ installAppWithPath() { # $1: path to app to install in $targetDir
                 printlog "notifying"
                 displaynotification "$message" "No update for $name!"
             fi
-            cleanupAndExit 0 "No new version to install" WARN
+            cleanupAndExit 0 "No new version to install" REG
         else
             printlog "Using force to install anyway."
         fi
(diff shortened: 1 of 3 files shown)


fragments/functions.sh
```diff
@@ -497,16 +497,21 @@ installAppWithPath() { # $1: path to app to install in $targetDir
 
         # copy app to /Applications
         printlog "Copy $appPath to $targetDir"
-        if ! ditto "$appPath" "$targetDir/$appName"; then
-            cleanupAndExit 7 "Error while copying" ERROR
+        copyAppOut=$(ditto -v "$appPath" "$targetDir/$appName" 2>&1)
+        copyAppStatus=$(echo $?)
+        deduplicatelogs "$copyAppOut"
+        printlog "Debugging enabled, App copy output was:\n$logoutput" DEBUG
+        if [[ $appVerifyStatus -ne 0 ]] ; then
```
scriptingosx, line 504: shouldn't this be checking `copyAppStatus`?

Theile: Did I copy-paste and not correct it all? Have to look further into this, but i guess you are right.

PR #430, opened 2022-03-03
Always download Chrome universal

Maarten-Wijnants:
Google doesn't seem to update the Chrome intel download fast enough. This may cause installomator to see there is a new version available, but still download an older version.

We can use the universal download for all platforms anyway.


scriptingosx:
we originally chose to not always download the universal because of the pkg and app size. However, if this avoids issues, rather than create more, then it would simplify

PR #434, opened 2022-03-03
Update functions.sh

Theile:
Fixed the `apples` line so it's now `$applist`. 

Also changed a few logging levels around this.


scriptingosx:
I failed at merging this into 9.1. moving to 9.2

Theile:
Ok, but should merge fine.

PR #437, opened 2022-03-05
Add swiftdialog

bartreardon:
added swiftdialog label (same as `dialog` but updated name)


Theile:
I will close this PR, as we have PR #435 that include the change, no matter if you call label as "dialog" or "swiftdialog"…

PR #454, opened 2022-03-22
Updated 'onlyofficedesktop'

adibue:
Added `downloadURL` for ARM.
Fixed `downloadURL` for x86_64 as well.
Unfortunately I could not find any other version source for `appNewVersion`.
Should work just fine tho.


scriptingosx:
thanks!

PR #464, opened 2022-04-04
Add aText Legacy which provides aText 2 for those with older licenses.

darklink87:
The link used for "atext" was updated a few months ago to include aText 3 which drastically changed their licensing model to be subscription based and does not accept aText 2 licenses. aText 2 is still supported until further notice, but is now called "aText Legacy" and available through a different link.


scriptingosx:
Next time, please branch your PR from `main` branch, not `dev` and create a new file for the label in `fragments/labels/` not in the root Installomator.sh. See [CONTRIBUTING](https://github.com/Installomator/Installomator/blob/main/CONTRIBUTING.md)

That said, thanks for contributing!

PR #465, opened 2022-04-05
Add latest version check to firefoxpkg and gpgsuite

Maarten-Wijnants:
Add the appNewVersion check to the firefoxpkg and gpgsuite label so it only installs updates if there is a newer version available.


scriptingosx:
Thanks!

PR #471, opened 2022-04-05
Added install logic for BlueJeans Events - "bluejeansevents"

gknackstedt:
Added install logic for currently available version of BlueJeans Events client by Verizon. This application only has one .pkg URL currently and does not have architecture specific download URLs.

Added as "bluejeansevents" after existing logic for "bluejeans", beginning on like 1720


gknackstedt:
wrong branch

PR #489, opened 2022-04-14
Add IGNORE_UPDATETOOL option

Maarten-Wijnants:
This option allows you to ignore the update tool when it is defined in a label like googlechromepkg.

It can be useful if you want to make sure updates are applied and don't want to depend on the end user restarting the application.


scriptingosx:
wouldn't `INSTALL=force` have the same effect?

mwmatter:
I'll chime in on this to give my two cents. I'd love an option to ignore the update tool as well. I think a new argument to the INSTALL variable works well.

Maarten-Wijnants:
I solved my original issue with Chrome by using a configuration profile that forces the user to restart Chrome if they haven't applied the updates after 2 days.
However I still think an option to ignore the update tool still could be a useful addition as we've had some problems with Installomator in combination with Microsoft Autoupdate lately. I'll see if it can be added to the INSTALL parameter.

Maarten-Wijnants:
I've created a new pull request with the INSTALL=ignore-update-tool option.
https://github.com/Installomator/Installomator/pull/899

PR #492, opened 2022-04-18
updates for ringcentral app

lart2150:
* update installers to install latest version
 * add arm installer (no universal installer yet)

Ringcentral renamed their chat app from glip to ringcentral, a few years ago they renamed the .app file to ringcentral.  Recently they also added a arm build of their chat app.


scriptingosx:
Thank you!

PR #493, opened 2022-04-21
Add simple, no prerequisite install script installInstallomator.sh

acodega:
It would be useful to have a script available to perform an Installomator install in a platform/MDM agnostic way with no app installs after. The script also can be used as a function in a larger deployment script.

### DIFF
--- a/MDM/InstallInstallomator.sh
+++ b/MDM/InstallInstallomator.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # This script is meant to povide the simplest MDM/management platform agnostic way to install Installomator
 # The only requirement is an Internet connection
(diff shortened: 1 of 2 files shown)


MDM/InstallInstallomator.sh
```diff
@@ -0,0 +1,43 @@
...
+  else
+    echo "Package verification failed before package installation could start. Download link may be invalid. Aborting."
+    exitCode=1
+    exit $exitCode
+  fi
+  # Remove the temporary working directory when done
+  echo "Deleting working directory '$tempDirectory' and its contents"
+  rm -Rf "$tempDirectory"
+else
+  echo "Installomator already installed."
```
scriptingosx, line 40: Should we run `Installomator.sh installomator` here to make sure it is on the latest version, or rather leave that to the admin later.

acodega: I think it's outside of the scope of the script.. I don't think this script also needs to handle upgrading.

PR #494, opened 2022-04-21
Fix for missing comment in header

sidbena:
Add a # to verify $CLIInstaller and $CLIArguments are executed.
Also small fix to indentation.


scriptingosx:
Thanks!

PR #498, opened 2022-04-26
Update snagit

marcelpogorzelski:
Yearly Snagit update. Added `appNewVersion`

### DIFF
--- a/fragments/labels/snagit.sh
+++ b/fragments/labels/snagit.sh
@@ -5,5 +5,6 @@ snagit2020)
     name="Snagit 2022"
     type="dmg"
     downloadURL="https://download.techsmith.com/snagitmac/releases/Snagit.dmg"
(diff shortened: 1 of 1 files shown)


scriptingosx:
I don't like that we are adding a new label for each year. But now that we have established it the past, I am afraid we are stuck with it? Or can we just remove all `snagit20xx`?

marcelpogorzelski:
For my part you can remove them. I only need the newest. I'll try to find some time tomorrow to make one for each version.

marcelpogorzelski:
I added one label for each year. Made 2022 download URL generated the same way as the rest so we get easy transfer to inevitable 2023 release.

scriptingosx:
Thank you!

scriptingosx:
(note the label for snagit2021 was wrong, but I fixed it in the merge)

PR #499, opened 2022-04-26
Update Camtasia 2021

marcelpogorzelski:
Fixed `name` and added `appNewVersion`

### DIFF
--- a/fragments/labels/camtasia.sh
+++ b/fragments/labels/camtasia.sh
@@ -1,6 +1,8 @@
-camtasia)
-    name="Camtasia 2020"
+camtasia|\
+camtasia2021)
+    name="Camtasia 2021"
     type="dmg"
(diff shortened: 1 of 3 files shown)


scriptingosx:
I am not familiar with app. Is Camtasia 2020 still available for download? Is it being updated independently? Are there reasons (licensing) people might want to _keep_ the 2020 version and _not_ update? Or this merely decorative change of the name?

marcelpogorzelski:
I am unfortunately not familiar myself. I just updated it to the newest version and fixed the appNewVersion. Quick check on their website shows that older version are available.

marcelpogorzelski:
I checked. It should be possible to make one for each version.I don't know if it's really needed.

marcelpogorzelski:
Sorry for the new line blunder. FIxed it now

PR #505, opened 2022-05-02
Update installFromPKG Funciton

theadamcraig:
Adding functionality for a choiceChangesXML variable to allow a customized .pkg installer
this functionality will be used with the bluejeanswithaudiodriver application label.


scriptingosx:
I like the idea in general, but passing the choiceChanges XML data as an additional argument/variable seems clunky and error prone. These xmls can get fairly big.

PR #506, opened 2022-05-02
Create bluejeanswithaudiodriver label

theadamcraig:
This will work with Pull Request #505

and allow the BlueJeans package to be installed with the choiceChangesXML variable so that the audio driver is installed along with the software.


scriptingosx:
Thank you!

PR #508, opened 2022-05-03
Fixed Citrix Workspace appNewVersion

marcelpogorzelski:
`cut` didn't work for `appNewVersion` so I changed it to awk.


scriptingosx:
Thanks!

PR #514, opened 2022-05-04
1Password 8

micahflee:
1Password 8 was just released, and Installomator.sh only supports 1Password 7 so far. This adds support for the new version as well.


scriptingosx:
merging this for now. we can (and) should add the `appNewVersion` later

PR #515, opened 2022-05-05
Update Label: Grammarly

drtaru:
Grammarly Desktop ships with a DMG containing Grammarly Installer.app need to call the actual app to move to the Applications folder with the correct name.


scriptingosx:
Thanks

PR #531, opened 2022-05-13
Update README.md shields.io badges urls

FletcherS7:
Now pointed to Installomator/Installomator instead of scriptingosx/Installomator
Fixes Release and Downloads badges showing "no releases or repo found" error


scriptingosx:
Hah, should have noticed that. Thank you!

PR #533, opened 2022-05-17
add appNewVersion to egnyte label

### LINKED ISSUE #500: Egnyte keeps updating
Hi Installomator team!

We are using installomator with our MDM system, which runs each label daily. One label we deploy is Egnyte, which seems to want to try and update every day, even when there is not actually a new version released. It seems to be reading the installed version incorrectly.

Any assistance with sorting this would be much appreciated.

Cheers!

scriptingosx:


scriptingosx:
closes #500

Issue #467, opened 2022-04-05
Blender error when trying to download - exit code 2


adorador:
Hello,

It looks like with the recent release of Blender version 3.1.2 there are two releases now for the Mac - one for Intel and one for Apple Silicon.  I believe this is what has broken the ability for Installomator to download the dmg.  Here is the log from my run:

```
Script exit code: 2
Script result: 2022-04-05 14:05:27 : REQ   :  : shifting arguments for Jamf
2022-04-05 14:05:27 : REQ   : blender : ################## Start Installomator v. 9.1, date 2022-03-18
2022-04-05 14:05:27 : INFO  : blender : ################## Version: 9.1
2022-04-05 14:05:27 : INFO  : blender : ################## Date: 2022-03-18
2022-04-05 14:05:27 : INFO  : blender : ################## blender
2022-04-05 14:05:28 : INFO  : blender : BLOCKING_PROCESS_ACTION=tell_user
2022-04-05 14:05:28 : INFO  : blender : NOTIFY=silent
2022-04-05 14:05:28 : INFO  : blender : LOGGING=INFO
2022-04-05 14:05:28 : INFO  : blender : LOGO=/System/Applications/App Store.app/Contents/Resources/AppIcon.icns
2022-04-05 14:05:28 : INFO  : blender : Label type: dmg
2022-04-05 14:05:28 : INFO  : blender : archiveName: blender.dmg
2022-04-05 14:05:28 : INFO  : blender : no blocking processes defined, using blender as default
2022-04-05 14:05:28 : INFO  : blender : App(s) found: /Applications/blender.app
2022-04-05 14:05:28 : INFO  : blender : found app at /Applications/blender.app, version 3.1.0, on versionKey CFBundleShortVersionString
2022-04-05 14:05:28 : INFO  : blender : appversion: 3.1.0
2022-04-05 14:05:28 : INFO  : blender : Latest version not specified.
2022-04-05 14:05:28 : REQ   : blender : Downloading  to blender.dmg
2022-04-05 14:05:28 : ERROR : blender : error downloading 
ls: blender.dmg: No such file or directory
2022-04-05 14:05:28 : ERROR : blender : File list: 
2022-04-05 14:05:28 : ERROR : blender : File type: blender.dmg: cannot open `blender.dmg' (No such file or directory)
2022-04-05 14:05:28 : INFO  : blender : App not closed, so no reopen.
2022-04-05 14:05:28 : ERROR : blender : ERROR: Error downloading  error:
[...]

PR #534, opened 2022-05-17
Fix Blender Download

cardinaldeville:
All credit goes to this issue: https://github.com/Installomator/Installomator/issues/467


cardinaldeville:
whoops. edited the full script. will submit a diff pull request.

PR #563, opened 2022-05-26
Update Google Drive Label

FletcherS7:
Package ID is longer correct for the current version of Google Drive.

Update Package ID and add app name because it differs from the label


scriptingosx:
Thanks

PR #567, opened 2022-06-02
Fix downloadURL for Sublime Text

mbulmer-addepar:
Download URL changed and the curl command was returning more than one result, causing the script to exit.
- Updated downloadURL
- Limited curl output to first result only


scriptingosx:
duplicate, closed with #593

Issue #573, opened 2022-06-10
Change app quitting/killing behaviour 


ThomasKaiser:
Are you open for one or more PRs to improve/change the following behaviour?

  * unified quitting behaviour: two quit attempts are made followed by 3 attempts to kill the process nicely (SIGTERM) and if this doesn't work hard (SIGKILL), see function `QuitOrKillGently` below
  * new property `LaunchDaemonsToUnload` to unload/reload LaunchDaemons that interfere with installation, see function `DealWithLaunchDaemon` below
  * I18N/Localization
  * maybe refactor structure to call a `main` function and allow the script to be sourced for easier debugging

Current app quitting behaviour is IMO quite annoying having to wait multitudes of 30 seconds instead of _always_ quitting the most polite way possible (`quit` event, followed by `SIGTERM`, followed by `SIGKILL`). Same with constantly localizing our local Installomator copy.

    QuitOrKillGently() {
    	# function that gets called with process name as $1
    	printlog "telling app \"$1\" to quit"
    	runAsUser osascript -e "tell app \"$1\" to quit"
    	sleep 5
    	if pgrep -xq "$1"; then
    		runAsUser osascript -e "tell app \"$1\" to quit"
    		sleep 5
    	fi
    
    	# walk through all processes that can be found using pgrep and first send them a
    	# SIGTERM and after 3 seconds a SIGKILL
    	RemainingPIDs=($(pgrep "$1"))
    	Iteration=0
    	while [ ${#RemainingPIDs[@]} -gt 0 -a ${Iteration} -lt 3 ] ; do
    		for PID in ${RemainingPIDs} ; do
    			Process="$(ps ${PID} | awk -F" /" "/${PID}/ {print \"/\"\$2}")"
    			printlog "sending SIGTERM to PID ${PID}: ${Process}"
    			kill ${PID}
    		done
    		sleep 5
    		RemainingPIDs=($(pgrep "$1"))
    		for PID in ${RemainingPIDs} ; do
    			Process="$(ps ${PID} | awk -F" /" "/${PID}/ {print \"/\"\$2}")"
    			printlog "sending SIGKILL to PID ${PID}: ${Process}"
    			kill -9 ${PID}
    		done
    		sleep 3
    		RemainingPIDs=($(pgrep "$1"))
    		((Iteration++))
    	done
    } # QuitOrKillGently
    
    DealWithLaunchDaemon() {
    	# function that stops/starts launchdaemons that could otherwise interfere with install
[...]

PR #574, opened 2022-06-10
Change app quitting/killing behaviour 

ThomasKaiser:
(continuing #573)

Change app quitting/killing behaviour and additionally invent new `$LaunchDaemonsToUnload` variable that will be handled by `DealWithLaunchDaemon` function.

The latter is a new addition with zero consequences since not changing anything. The purpose is to define background tasks that might interfere with installations (at our place we have a daemon watching every minute whether a VoIP application is running and restarts if needed)

The more serious or 'disruptive' change is introducing the `QuitOrKillGently` to quit and kill apps in a different way than before. With the new functionality per string in `$blockingProcesses` the following happens:

  * send a `quit` event, wait 5 seconds, recheck using `pgrep -xq`, send another `quit` event, wait another 5 seconds
  * create a list of matching PIDs using `pgrep`, walk through them and send a `SIGTERM` signal to each, wait 5 seconds
  * again create a list of matching PIDs using `pgrep`, walk through them and send a `SIGKILL` signal to each, wait 3 seconds
  * iterate through the two steps above for max 3 times and return in any case
  * if there are remaining processes the `checkRunningProcesses` function will catch them via `pgrep -xq` and exit as before

The purpose of this 3 step attempt to quit an app as gently as possible while also saving some time (not waiting multiple times 30 seconds for nothing).

But unfortunately this change is disruptive in the following way.

The old way `checkRunningProcesses` worked was partially broken if processes were immediately restarted. One such example is `1PasswordNativeMessageHost` as member of `$blockingProcesses` for both `1password7` and `1password8`. You can quit or kill `1PasswordNativeMessageHost` in any possible way but to no avail since immediately restarted with a new PID.

Quoting the log:

    found blocking process 1PasswordNativeMessageHost
    telling app "1PasswordNativeMessageHost" to quit
    sending SIGTERM to PID 11166: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost
    sending SIGKILL to PID 12263: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost
    sending SIGTERM to PID 12273: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost
    sending SIGKILL to PID 12284: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost
    sending SIGTERM to PID 12294: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost
    sending SIGKILL to PID 12305: /Applications/1Password 7.app/Contents/Library/LoginItems/1Password Extension Helper.app/Contents/MacOS/1PasswordNativeMessageHost

And then this starts over and over again. The only way to 'fix' this is either removing the task in question from `$blockingProcesses` in the app definition (safe since it never worked before) or to take special precautins and onload the LaunchAgent first (this will be part of another PR to deal not only with LaunchDaemons but also LaunchAgents):

    launchctl asuser launchctl stop 2BUA8C4S2C.com.agilebits.onepassword7-helper
    launchctl asuser launchctl remove 2BUA8C4S2C.com.agilebits.onepassword7-helper

Only after unloading the LaunchDaemon the corresponding process `1PasswordNativeMessageHost` can really be stopped. In this case the solution is simply (since stopping never really worked removing the string from the `$blockingProcesses` list) but this could be disruptive for other apps as well.

I searched through the current definitions and found the following entries that might be candidates for troubles:

    blockingProcesses=( "Autodesk Fusion 360" "Fusion 360" )
    blockingProcesses=( "Google Docs" "Google Drive" "Google Sheets" "Google Slides" )
    blockingProcesses=( "iStat Menus" "iStatMenusAgent" "iStat Menus Status" )
    blockingProcesses=( "Keyboard Maestro Engine" "Keyboard Maestro" )
    blockingProcesses=( "Mattermost Helper.app" "Mattermost Helper (Renderer).app" "Mattermost Helper (GPU).app" "Mattermost Helper (Plugin).app" )
    blockingProcesses=( "Microsoft AutoUpdate" "Microsoft Word" "Microsoft PowerPoint" "Microsoft Excel" "Microsoft OneNote" "Microsoft Outlook" "OneDrive" )
    blockingProcesses=( "Microsoft AutoUpdate" "Microsoft Word" "Microsoft PowerPoint" "Microsoft Excel" "Microsoft OneNote" "Microsoft Outlook" "OneDrive" "Teams")
    blockingProcesses=( NVivo NVivoHelper )

So without testing with the corresponding apps this PR might introduce failing updates if one of those strings above is also the result of a process being fired up (and restarted immediately) by a LaunchDaemon or LaunchAgent.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -1727,7 +1727,6 @@ axurerp10)
     expectedTeamID="HUMW6UU796"
     versionKey="CFBundleVersion"
     appName="Axure RP 10.app"
-    blockingProcesses=( "Axure RP 10" )
     ;;
 balenaetcher)
     name="balenaEtcher"
@@ -1847,7 +1846,6 @@ caffeine)
     downloadURL=$(downloadURLFromGit IntelliScape caffeine)
     appNewVersion=$(versionFromGit IntelliScape caffeine)
     expectedTeamID="YD6LEYT6WZ"
-    blockingProcesses=( Caffeine )
     ;;
 cakebrew)
     name="Cakebrew"
@@ -2204,7 +2202,6 @@ drawio)
     downloadURL="$(downloadURLFromGit jgraph drawio-desktop)"
     appNewVersion="$(versionFromGit jgraph drawio-desktop)"
     expectedTeamID="UZEUFB4N53"
-    blockingProcesses=( draw.io )
     ;;
 drift)
     # credit Elena Ackley (@elenaelago)
@@ -2261,7 +2258,6 @@ egnytewebedit)
     appName="Egnyte WebEdit.app"
     blockingProcesses=( NONE )
     ;;
-
 element)
     name="Element"
     type="dmg"
@@ -2501,7 +2497,6 @@ flux)
     downloadURL="https://justgetflux.com/mac/Flux.zip"
     expectedTeamID="VZKSA7H9J9"
     ;;
-
 flycut)
     name="Flycut"
     type="zip"
@@ -3153,9 +3148,7 @@ linear)
     expectedTeamID="7VZ2S3V9RV"
     versionKey="CFBundleShortVersionString"
     appName="Linear.app"
-    blockingProcesses=( "Linear" )
     ;;
-
 logioptions|\
 logitechoptions)
     name="Logi Options"
@@ -3337,7 +3330,6 @@ azuredatastudio)
     appNewVersion=$(versionFromGit microsoft azuredatastudio )
     expectedTeamID="UBF8T346G9"
     appName="Azure Data Studio.app"
-    blockingProcesses=( "Azure Data Studio" )
     ;;
 microsoftazurestorageexplorer)
     name="Microsoft Azure Storage Explorer"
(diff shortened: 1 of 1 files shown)


ThomasKaiser:
After adding `LaunchAgentsToStop functionality` for testing purposes the script over there https://github.com/ThomasKaiser/Installomator/blob/main/Installomator.sh is ready for testing.

Not planning to do any further updates but waiting for feedback now.

ThomasKaiser:
Oops, just realized that a _real_ PR needs to go against https://github.com/Installomator/Installomator/blob/main/fragments/functions.sh – apologies for that.

But maybe we can discuss functionality still here and editing the assembled script may help with testing?

scriptingosx:
While I can see the desire to do this, shouldn't the vendor's installer take care of that? I believe there are too many variables here to build a general solution?

PR #577, opened 2022-06-14
Add blocking process checking to update tool function

FletcherS7:
Currently when a defined update tool is used or the label type is updateronly, Blocking Processes are not checked.
Looks like this is because the Update tool function is above the blocking processes function in main.sh.

To get this working I copied the blocking process action function into the update tool function.
This may not be the best way to do this.

This also might cause problems if blocking processes only need to be checked when not using an update tool.

### DIFF
--- a/fragments/main.sh
+++ b/fragments/main.sh
@@ -169,6 +169,17 @@ fi
 if [[ (-n $appversion && -n "$updateTool") || "$type" == "updateronly" ]]; then
     printlog "appversion & updateTool"
     if [[ $DEBUG -ne 1 ]]; then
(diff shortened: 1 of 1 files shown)


scriptingosx:
I believe this might be an intentional choice as the thought is that the update tool might/should have it's own logic to deal with running processes.

@Theile do you hav comment on this?

Theile:
My best memory of this is that it's correct what Armin is writing. 

If the software should only be updated it will use `updateTool` to do this, and that variable will point to the tool within the software itself so we expect that to handle the situation.

Installomator will not download anything in this situation, so we didn't think that Installomator should check for anything in these cases.

PR #586, opened 2022-06-20
Disable Community Patch version URL for AWS VPN Client

FletcherS7:
Disabled appNewVersion URL that was using beta2.communitypatch.com since Community Patch was shutdown.


scriptingosx:
Thank you!

PR #602, opened 2022-06-29
Update gotomeeting.sh

jshtux:
get the appNewVersion for GoTo Meeting


jshtux:
need to also include build number to get full version match from CFBundleShortVersionString

PR #606, opened 2022-07-04
Update vlc.sh

macmangr:
Changed appNewVersion check, so that no download is started if the same version of the software is installed

Changed download url


Theile:
Thank you, but another PR is also fixing this label, so I close this.

PR #615, opened 2022-07-07
Edited Google Drive label to be correct. It was searching for Google …

EddieV2:
…Drive File Stream.app instead of Google Drive.app. Also updated appNewVersion, but commented it out due to latest version not being available for download (will uncomment once workaround is found). Also added in correct PackageID for appropriate architecture.


FletcherS7:
This was fixed in https://github.com/Installomator/Installomator/pull/563

PR #628, opened 2022-07-21
Update to Alfred 5

bp88:
"downloadURL" variable regex pulls down the latest version of Alfred 5. Updated the "appName" variable to reflect the new app bundle name of "Alfred 5.app".


Theile:
You should make your changes to the files below the "fragments" folder. In there you find "labels"-folder, and then you can find each label to be an individual file. Those are the ones we need changed. The. full Installomator.sh script will be overwritten the next time we build a new version.

Please correct this, thank you.

PR #629, opened 2022-07-23
Add ARM version of Miro

vdmchkv:
Recently Miro published ARM version of their desktop application. This changes enables download of a native ARM application on capable MacBook's.


vdmchkv:
main branch already contains this patch

PR #635, opened 2022-07-29
New Label: Mendeley Desktop

kryptonit1:


Theile:
What about `appNewVersion` for mendeleydesktop?

kryptonit1:
I just noticed that Mendeley Desktop will be replaced by Mendeley Reference Manager soon:

https://blog.mendeley.com/2022/02/22/introducing-mendeley-reference-manager-designed-for-todays-researcher-workflow/

So I've created a new label for Mendeley Reference Manager that includes appNewVersion as well instead of this label: #713

PR #637, opened 2022-08-01
Display version info of application in notification windows for prompts

dnikles:
When a user is prompted to update an application, it may be helpful to show them the version of the current installed application, and the version of the available update. This may help influence those that just choose "not now" every time, because they may see that they are X versions behind. If we do not know the value of $appNewVersion, this additional information is not displayed.


dnikles:
Example:
slack update
Example where $appNewVersion is unknown:
packages

PR #641, opened 2022-08-11
swiftDialog Integration

scriptingosx:
Installomator will now write commands with  progress updates for download and installation (with pkgs) to a swiftDialog command file.


scriptingosx:
implements #632

MDM/Jamf/00_Prepare_SwiftDialog.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# display first screen
+open -a "$dialogApp" --args \
+        --title none \
+        --icon "$icon" \
+        --message "$message" \
+        --mini \
+        --progress 100 \
+        --position bottomright \
+        --ontop \
```
bartreardon, line 65: one issue with `--ontop` is that it will force dialog to be the focused app every time it receives an update. This has the effect that you can't do anything while the process is running. e.g. if it was a larger download and install this could become disruptive.

MDM/Jamf/zz_Quit_SwiftDialog.sh
```diff
@@ -0,0 +1,54 @@
...
+dialogUpdate "progresstext: Done"
+
+# pause a moment
+sleep 0.5
+
+dialogUpdate "quit:"
+
+# just to be safe
+killall "Dialog"
+
```
bartreardon, line 51: I think the killall is running before dialog gets a chance to quit by itself. This will leave an orphaned `tail -f` process but I might re-check that as I had to change some of the logic around that for the issue Brad reported and the `quit:` command isn't cleaning up. if you've done a bit of testing you probably have a few instances of `tail` running

scriptingosx: I actually do see the tail processes. So remove the `killall`?

MDM/Jamf/00_Prepare_SwiftDialog.sh
scriptingosx: https://github.com/Installomator/Installomator/pull/641/commits/614da59311c8e6923eb3c80e4bdd346f959183f5

MDM/Jamf/zz_Quit_SwiftDialog.sh
scriptingosx: inserted a `sleep 0.5` https://github.com/Installomator/Installomator/pull/641/commits/ef19aa28fe48528b9677992bccca10d145d9afe6

fragments/arguments.sh
```diff
@@ -99,6 +100,16 @@ if [[ "$(whoami)" != "root" && "$DEBUG" -eq 0 ]]; then
     cleanupAndExit 6 "not running as root, exiting" ERROR
 fi
 
+
+# check Swift Dialog presence and version
+DIALOG_CMD="/usr/local/bin/dialog"
```
Theile, line 105: Will this not override a custom value in the beginning of the script or set manually as an argument to the script call?

scriptingosx: This is `DIALOG_CMD` not `DIALOG_CMD_FILE`...

PR #643, opened 2022-08-12
updated redundant exit codes

### LINKED ISSUE #561: Create a unique exit code for "App previously installed from App Store, and we respect that"
Currently "App previously installed from App Store, and we respect that" exits with a code of 1. Other code 1 exits are "hard fails" like failure to access directories, running an outdated/unsupported version of macOS, providing an unknown label.
Having a unique exit code for this specific error would be useful for scripting/self-service. I see this as more of a "soft fail" like "this is already handled by VPP/App Store" and not a "you have a problem or misconfiguration." I'd like to be able to differentiate based on the Installomator.sh exit code so that my wrapper script can treat this differently from a misconfiguration. Thank you!

scriptingosx:
See also: https://github.com/Installomator/Installomator/wiki/Installomator-Exit-Codes


scriptingosx:
closes #561

PR #665, opened 2022-08-23
label screamingfrogspider.sh fix, dmg per arch

niavlysAO:
ScreamingFrogSpider now has different dmg per CPU arch.
Download page provides 2 dmg links, therefore we must fix downloadURL grep


Theile:
Thank you!

PR #678, opened 2022-09-02
Update acroniscyberprotectconnectagent.sh

Theile:
`blockingProcesses=( NONE )`


Theile:
I never managed to find the `appNewVersion`. No version on web site, and could not find any rss-feeds or anything.

PR #679, opened 2022-09-02
Add Dockutil Support

drtaru:
This adds support for dockutil to auto add apps to the dock, note the dock will restart when set to yes.
to use set ADDTODOCK to yes or pass ADDTODOCK=yes as an argument


fragments/functions.sh
```diff
@@ -1009,3 +1011,28 @@ updateDialog() {
...
+    if [ ! -e "/usr/local/bin/dockutil" ]; then
+        printlog "Dockutil missing, cannot add $appName to dock"
+        return
+    fi
+    if [[ -d "$installedAppPath" ]]; then
+        printlog "Adding $appName to the dock."
+        loggedInUserDock="/Users/$currentUser/Library/Preferences/com.apple.dock.plist"
+        /usr/local/bin/dockutil --add $installedAppPath/ $loggedInUserDock &> /dev/null
+    else
+        printlog "$appNmae is a binary, will not add to dock"
```
acodega, line 1036: Typo in `$appName`

fragments/header.sh
```diff
@@ -157,6 +157,10 @@ DIALOG_LIST_ITEM_NAME=""
 # listitem.
 # When the variable is unset, progress will be sent to Swift Dialog's main progress bar.
 
+# Add resulting App to the Dock using dockutil
```
acodega, line 160: Lets call it `Add installed app` rather than resulting app, and describe that the app will be added at the end of the user's current dock. Can we also add a line that dockutil has an Installomator label. Perhaps remove the dockutil GitHub URL and just reference that a dockutil label exits.

fragments/header.sh
```diff
@@ -157,6 +157,10 @@ DIALOG_LIST_ITEM_NAME=""
 # listitem.
 # When the variable is unset, progress will be sent to Swift Dialog's main progress bar.
 
+# Add installed App to the end of existing user's using dockutil
+ADDTODOCK=""
```
scriptingosx, line 161: Standard format for user-settable variables in Installomator is capitalized snake-case, so please change this to `ADD_TO_DOCK`

scriptingosx:
One more question: what happens if you have `ADD_TO_DOCK` enabled and run installomator multiple times? will it add multiple icons for the same app?

drtaru:
If ADD_TO_DOCK is enabled but the item already exists dockutil fails silently and does not modify the dock. 
I don't see much use trying to log that outcome though as the end result of the icon being in the dock is still true.
I will modify the help text to reflect this.

Theile:
I would think that this should be moved to a separate script. That way we can keep Installomator free from 3. party software requirements.

Actually this suggestion was an inspiration for me to add this functionality to the MDM scripts.

PR #685, opened 2022-09-12
New Label: MacTeX

kryptonit1:


Theile:
What about `appNewVersion`, it's an important variable? What have you found for that?

kryptonit1:
> What about `appNewVersion`, it's an important variable? What have you found for that?

The problem is that this is a bundle/distribution of an application called "TeX Live Utility" and other Mac specific utilities. The distribution itself changes version based on release year (MacTeX 2021, MacTeX 2022 etc), but this does not match the version of the TeX Live Utility application installed for example.

https://www.tug.org/mactex/
https://www.tug.org/mactex/mactex-download.html
https://www.tug.org/mactex/updateschedule.html

So I don't know how to solve appNewVersion for this bundle/distribution.

scriptingosx:
Thank you!

PR #687, opened 2022-09-12
Swap Adobe URL parsing filter

acodega:
The filters for ARM and Intel were swapped by accident, this correctly returns the ARM URL for ARM and Intel URL for Intel.


Theile:
Looks to me, like this is right. It installed Adobe CC on my Apple Silicon Mac.

PR #691, opened 2022-09-14
If GitHub does not give an URL we can fall back on API call

Theile:
Today something strange happened for Installomator on GitHub. Or latest release would spend too much time loading the assets, and would not return an URL, like here:

```
% gitusername="Installomator"; gitreponame="Installomator"; curl -sfL "https://github.com/$gitusername/$gitreponame/releases/latest" | tr '"' "\n" | grep -i "^/.*\/releases\/download\/.*\.pkg" | head -1
% gitusername="bartreardon"; gitreponame="swiftDialog"; curl -sfL "https://github.com/$gitusername/$gitreponame/releases/latest" | tr '"' "\n" | grep -i "^/.*\/releases\/download\/.*\.pkg" | head -1
/bartreardon/swiftDialog/releases/download/v1.11.2/dialog-1.11.2-3095.pkg
```

As can be seen above it worked for swiftDialog, but not for Installomator.

Now we can fall back on the API call if we do not get the URL in first try.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -162,13 +162,15 @@ downloadURLFromGit() { # $1 git user name, $2 git repo name
     fi
 
     if [ -n "$archiveName" ]; then
-        downloadURL=https://github.com$(curl -sfL "https://github.com/$gitusername/$gitreponame/releases/latest" | tr '"' "\n" | grep -i "^/.*\/releases\/download\/.*$archiveName" | head -1)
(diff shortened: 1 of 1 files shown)


Theile:
@scriptingosx this fixes the GitHub download (when not using API). I can see how Jamf Pro is distributing profile distribution, but other solutions handles push of commands at specific times, rather than a week ago. You can still choose to put API call first, if that is the safe bet, but falling back increases the download reliability.

PR #708, opened 2022-09-30
Update audacity label to check for arch

dnikles:
previous version of label was downloading the arm64 version every time


Theile:
Thank you!

PR #716, opened 2022-10-06
IBM Notifier

Theile:
IBM for some reason is adding too much to the “tag” on Github, so they have “Version 2.9.1 Build 96” for the current latest version. That ends up being read as 2.9.196, and not only 2.9.1 that the app version is.

I upload this, so we have it ready, but not sure if `appNewVersion` should stay or be commented out.


scriptingosx:
Thank you!

PR #724, opened 2022-10-11
Fixing `adobereaderdc` and `adobereaderdc-install`

Theile:
So now we have a label for `adobereaderdc`, that previously also was named `adobereaderdc-install`, but the latter got it's own label, so now it should work.


scriptingosx:
Thank you!

PR #730, opened 2022-10-14
Camtasia 2022 update

marcelpogorzelski:
Application name for Camtasia is renamed each year. Maybe removing the standalone **camtasia** label and only leave **camtasia20XX** labels is a good idea?

An other idea would be doing a _curl_ for the **name** field to get the newest application name. Are you okay with dynamic application name?


scriptingosx:
Thank you!

PR #731, opened 2022-10-14
citrixworkspace: Update parsing logic for download and version URLs

sidbena:
I've updated the logic for parsing download and version URLs to use `xpath` instead of parsing html with `sed`.


scriptingosx:
Thank you!

PR #739, opened 2022-10-24
Update googleadseditor.sh

Theile:
I figured out `appNewVersion` but the app installed does not have a version to compare it to, so it is commented out.


scriptingosx:
Thank you!

Issue #518, opened 2022-05-10
New application installs have "update" messaging versus "installation complete" messaging.


mwmatter:
When installing a net new application on a system the notifications that appear act as if the app is an update versus new. This seems to be happening because the function to check the app version is run in the finishing function. Running the app version check is what detects if the application is an update by seeing if it already exits but the app will always exist by time you get to the finishing function since it just when through its install process. Instead it seems that there should be a separate function specifically for checking if the application already exists on the system which then sets the updateDetected variable. Within the getAppVersion function the updateDetected variable should be removed entirely.

PR #756, opened 2022-11-04
Updated finishing function for better detection of new install versus an update.

mwmatter:
Made some small changes to the finishing function that removes the `getAppVersion` call (this is what accidentally always sets the `updateDetected` variable to "YES". By instead removing this call and also using the `appNewVersion ` variable we get better results on a new install versus an update.


mwmatter:
This is related to issue: https://github.com/Installomator/Installomator/issues/518

Issue #379, opened 2022-01-14
Microsoft Visual Studio Code and the blockingProcess


adibue:
Hello, fellow MacSysAdmins.

I have the problem with **Visual Studio Code** that no blocking process is found when `blockingProcesses` is set to `( Code )`. However, it works fine when set to `( Electron )`, which is of course very unsexy as other running applications might also use Electron as their process.

Do you have the same problem? Is there perhaps a better way to correctly identify a running VS code in Installomator?

acodega:
I thought we had an open Issue for this, we've run in to this before, particularly with Visual Studio Code. The way we check running processes right now is too strict, however I haven't had time/knowledge to think through how we can improve it without then causing false positives. Removing -x, `pgrep -i Code`, which requires an exact match of the process name, returns a result when Visual Studio Code is running.

acodega:
I'm going to mark this as enhancement, it's not a bug since prep is working they way we've written it, but an improvement in process matching is needed. So far I haven't found other apps, including Slack or Teams, which are found when using `prep -ix Electron`. Maybe we should change the label to those processes, with a comment warning to test locally to avoid false positives.

acodega:
@mrazishere want to open a PR to change the blockingProcess to `Code Helper` ?

michael-zukrow-da:
could pgrep -fi "visual studio code.app/content" be used? seems more future proof

asri-tm:
i have since stopped using installomator for vs code deployment/patching, even with the `blockingProcesses` updated to Electron, I can never get the script to kill the process with the `prompt_user_then_kill` action without getting endless prompts.

modified [this](https://github.com/alectrona/install-update-app-in-dmg/blob/master/install-update-app-in-dmg.sh) to support zip archives and works perfectly for vs code deployment/patching

Single prompt, kills process, proceed to update and relaunch app.

Issue #375, opened 2022-01-13
Adobe Acrobat Reader cannot be quit for update


fleish:
This is a follow-up from #29 

While the addition of:
`blockingProcesses=( AdobeReader )`

successfully detects a running copy of Adobe Acrobat Reader DC, attempts to use Apple Script to tell said application to quit fail because it uses the blockingProcess name:
`+checkRunningProcesses:71> runAsUser osascript -e 'tell app "AdobeReader" to quit'`

The correct application name that needs to be passed in is "Acrobat Reader":
`osascript -e 'tell app "Acrobat Reader" to quit'`

I'm not sure how to reconcile the need to detect "AdobeReader" but tell "Acrobat Reader" to be quit instead.

Theile:
We can easily change the label blocking line to this:
```
blockingProcesses=( "Adobe Reader" )
```
But I need somebody to confirm that is the needed change.

fleish:
I'm afraid that change won't work as it keeps the script from detecting the app is currently running. I think the process that needs to be detected is `AdobeReader` but the app name that needs to be passed to osascript is `Acrobat Reader`

PR #757, opened 2022-11-04
Add functionality when finding and closing blocking processes to find the appropriate process name

sculliga:
There are issues with some apps where the process name found using `pgrep` differs from the app name when calling the apple script to tell the application to close.

Some examples of existing issues this should fix. #379 and #375 

To address this, I've added the following.

- Search for the process using both `pgrep` and `lsappinfo`. `lsappinfo` is used to match the app name as it's seen in Activity Monitor and for the apple script that tells the application to close
- Handle the 3 different scenarios for the process names
    - Process names match between `pgrep` and `lsappinfo`
    - Process was found via `pgrep`, and we need to find the `lsappinfo` process name for the apple script
    - Process was found via `lsappinfo`, and we need to find the `pgrep` process name to kill the process

We will populate the `processQuitName` and `processKillName` variables with correct values based on how the process was found and should handle every situation without any changes to labels.

I specifically wrote this due to an issue with closing Microsoft Teams, but this will address other applications in the issues above, like VSCode.


scriptingosx:
can you explain why we can't just add the alternative name to the list of blocking progress names?

scriptingosx:
see question to #758

sculliga:
As far as I understand it and what I have found in my testing, for the example of Microsoft Teams, when `pgrep` is run, the process name that matches is `Teams` but when calling the apple script `tell app XXX to quit`, `Teams` does not work. The correct name is `Microsoft Teams` which matches what Activity Monitor shows as the name. If we were to add `Microsoft Teams` as well to the `blockingProcesses` definition, `pgrep` wouldn't find `Microsoft Teams` and would just move on to the next item.

Due to needing both names in only some scenarios where the apple script app name is different, I added the alternative name definition for a couple of reasons

1. For an associative array, it needs to follow the `[key]=value` format so we cannot modify the existing `blockingProcesses` without breaking existing functionality.
2. The alternative name is not needed in most scenarios, so defining the `blockingProcessesAltName` is optional which leaves all existing labels functional but introduces the alternative name capability where needed.

sculliga:
Additionally, I updated the README file to include some documentation on the usage.

scriptingosx:
> If we were to add Microsoft Teams as well to the blockingProcesses definition, pgrep wouldn't find Microsoft Teams and would just move on to the next item.

How is that a problem?

sculliga:
> > If we were to add Microsoft Teams as well to the blockingProcesses definition, pgrep wouldn't find Microsoft Teams and would just move on to the next item.
> 
> How is that a problem?

@scriptingosx The only way the process can be closed is if `pgrep` is able to find the process running first, and then it performs the closing actions. Since `pgrep` will never see the application as Microsoft Teams, it would never proceed to the closing action.

`pgrep` needs to find the app as `Teams` and then to close it as `Microsoft Teams`

scriptingosx:
Thank you for bearing with me as I catch. I believe I understand the problem now.

An alternative here would be to not just use `pgrep` to determine if a process is running, but something that finds the app name which AppleScript recognizes, as well. I'd rather make the script "smarter" than add even more configuration burden to the admin. However, we might not be able to do that without requiring AppleScript event PPPC profile, which I'd also really want to avoid

sculliga:
@scriptingosx I agree with improving the script to be "smarter" instead of adding more configuration burden. I took some time and discovered a way to make the script "smarter" without the configuration changes while still addressing the issues. Please let me know if you have more questions about these changes.

uurazzle:
Something like this might resolve the issue?

`checkRunningProcesses() {
    # don't check in DEBUG mode 1
    if [[ $DEBUG -eq 1 ]]; then
        printlog "DEBUG mode 1, not checking for blocking processes" DEBUG
        return
    fi

    # Enhanced quit function that handles stubborn applications
    quitApplication() {
        local appName="$1"
        local maxAttempts=3
        local attempt=1
        
        printlog "Attempting to quit application: $appName"
        
        while [[ $attempt -le $maxAttempts ]] && pgrep -xq "$appName"; do
            printlog "Quit attempt $attempt of $maxAttempts for $appName"
            
            # Try AppleScript quit
            runAsUser osascript -e "tell app \"$appName\" to quit" 2>/dev/null || true
            
            # Wait progressively longer for each attempt
            local waitTime=$((10 * attempt))
            printlog "Waiting ${waitTime} seconds for $appName to quit"
            sleep $waitTime
            
            # Check if still running
            if pgrep -xq "$appName"; then
                printlog "$appName still running after AppleScript quit attempt $attempt"
                
                # On final attempt or if app is being stubborn, use more aggressive methods
                if [[ $attempt -eq $maxAttempts ]] || [[ $attempt -gt 1 ]]; then
                    printlog "Using more aggressive quit methods for $appName"
                    
                    # Try SIGTERM first (graceful termination)
                    pkill -TERM "$appName" 2>/dev/null || true
                    sleep 5
                    
                    # If still running, try SIGKILL (force kill)
                    if pgrep -xq "$appName"; then
                        printlog "Force killing $appName with SIGKILL"
                        pkill -KILL "$appName" 2>/dev/null || true
                        sleep 3
                    fi
                    
                    # Final cleanup - kill any related processes that might share the app name
                    if pgrep -f "$appName" > /dev/null 2>&1; then
                        printlog "Cleaning up related processes for $appName"
                        pkill -f "$appName" 2>/dev/null || true
                        sleep 2
                    fi
                fi
            else
                printlog "$appName successfully quit on attempt $attempt"
                break
            fi
            
            attempt=$((attempt + 1))
        done
        
        # Final verification
        if pgrep -xq "$appName"; then
            printlog "WARNING: $appName may still be running after all quit attempts"
            return 1
        else
            printlog "$appName successfully terminated"
            return 0
        fi
    }

    # try at most 3 times
    for i in {1..4}; do
        countedProcesses=0
        for x in ${blockingProcesses}; do
            if pgrep -xq "$x"; then
                printlog "found blocking process $x"
                appClosed=1

                case $BLOCKING_PROCESS_ACTION in
                    quit|quit_kill)
                        # Use enhanced quit function instead of simple AppleScript
                        quitApplication "$x"
                        
                        if [[ $i > 2 && $BLOCKING_PROCESS_ACTION = "quit_kill" ]]; then
                          printlog "Changing BLOCKING_PROCESS_ACTION to kill"
                          BLOCKING_PROCESS_ACTION=kill
                        else
                            # Reduced wait time since quitApplication handles its own timing
                            printlog "waiting 10 seconds before next process check"
                            sleep 10
                        fi
                        ;;
                    kill)
                      printlog "killing process $x"
                      # Use more comprehensive kill approach
                      pkill -TERM "$x" 2>/dev/null || true
                      sleep 3
                      if pgrep -xq "$x"; then
                          pkill -KILL "$x" 2>/dev/null || true
                          sleep 2
                      fi
                      # Clean up any related processes
                      pkill -f "$x" 2>/dev/null || true
                      ;;
                    prompt_user|prompt_user_then_kill)
                      button=$(displaydialog "Quit "$x" to continue updating? $([[ -n $appNewVersion ]] && echo "Version $appversion is installed, but version $appNewVersion is available.") (Leave this dialogue if you want to activate this update later)." "The application "$x" needs to be updated.")
                      if [[ $button = "Not Now" ]]; then
                        appClosed=0
                        cleanupAndExit 10 "user aborted update" ERROR
                      elif [[ $button = "" ]]; then
                        appClosed=0
                        cleanupAndExit 25 "timed out waiting for user response" ERROR
                      else
                        if [[ $BLOCKING_PROCESS_ACTION = "prompt_user_then_kill" ]]; then
                          # Use enhanced quit function
                          quitApplication "$x"
                          printlog "Changing BLOCKING_PROCESS_ACTION to kill"
                          BLOCKING_PROCESS_ACTION=kill
                        else
                          quitApplication "$x"
                        fi
                      fi
                      ;;
                    prompt_user_loop)
                      button=$(displaydialog "Quit "$x" to continue updating? $([[ -n $appNewVersion ]] && echo "Version $appversion is installed, but version $appNewVersion is available.") (Click "Not Now" to be asked in 1 hour, or leave this open until you are ready)." "The application "$x" needs to be updated.")
                      if [[ $button = "Not Now" ]]; then
                        if [[ $i < 2 ]]; then
                          printlog "user wants to wait an hour"
                          sleep 3600 # 3600 seconds is an hour
                        else
                          printlog "change of BLOCKING_PROCESS_ACTION to tell_user"
                          BLOCKING_PROCESS_ACTION=tell_user
                        fi
                      else
                        quitApplication "$x"
                      fi
                      ;;
                    tell_user|tell_user_then_kill)
                      button=$(displaydialogContinue "Quit "$x" to continue updating? (This is an important update). Wait for notification of update before launching app again." "The application "$x" needs to be updated.")
                      quitApplication "$x"
                      if [[ $i > 1 && $BLOCKING_PROCESS_ACTION = tell_user_then_kill ]]; then
                          printlog "Changing BLOCKING_PROCESS_ACTION to kill"
                          BLOCKING_PROCESS_ACTION=kill
                      fi
                      ;;
                    silent_fail)
                      appClosed=0
                      cleanupAndExit 12 "blocking process '$x' found, aborting" ERROR
                      ;;
                esac

                countedProcesses=$((countedProcesses + 1))
            fi
        done

        # Break out of retry loop if no processes are running
        if [[ $countedProcesses -eq 0 ]]; then
            printlog "All blocking processes have been terminated"
            break
        fi
        
        # Add a brief pause between retry attempts
        if [[ $i -lt 4 ]]; then
            printlog "Retry attempt $((i + 1)) in 5 seconds..."
            sleep 5
        fi
    done
}`

PR #758, opened 2022-11-04
Microsoft Teams Download URLs and new blocking process functionality

sculliga:
Using the current source of macadmins for the version number appears to be inconsistent with what version Microsoft is publically providing based on https://learn.microsoft.com/en-us/officeupdates/teams-app-versioning

I've changed the download URL to a static link that provides a link to the latest download of the macOS package and used the URL found to determine the `newAppVersion`

Additionally, Microsoft Teams does not close properly when you choose to quit the software. I created an additional PR to add functionality to fix that using the `blockingProcessesAltName` parameter. That PR is #757

### DIFF
--- a/fragments/labels/microsoftteams.sh
+++ b/fragments/labels/microsoftteams.sh
@@ -2,8 +2,8 @@ microsoftteams)
     name="Microsoft Teams"
     type="pkg"
     packageID="com.microsoft.teams"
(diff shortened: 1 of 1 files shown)


sculliga:
Closing based on changes in #757 and other changes to the Microsoft Teams label

PR #759, opened 2022-11-04
[Update] 1Password 8 Universal pkg

imfromthebay:
1Password released a universal pkg, editing the label to utilize it.


scriptingosx:
Thank you!

Issue #572, opened 2022-06-09
Deprecate Code42 label, no response from vendor about a static URL for download


Truezeldafan:
Currently, these links below from the C42 label:

https://download.code42.com/installs/agent/latest-mac.dmg
https://download.code42.com/installs/agent/latest-mac-arm64.dmg

Are not downloading the recently released version, 10.0. Code42 Representatives in the macadmins slack channel have stated the paths have been changed for the 10.0 release and there is currently not an updated "latest" link yet. They hope to provide one sometime.

naschenbrenner:
It would appear both of the "legacy" URL's are offline so this workflow is no longer working. Has there been any update on new "latest" URL's from Code42? 


![SCR-20221114-f76](https://user-images.githubusercontent.com/60478794/201719905-8a0d4452-283d-441e-81c3-70bace2cbf20.png)

acodega:
I think we're going to have to remove the label.

rchase254:
I tested the script locally using the following URL: https://download.crashplan.com/installs/agent/latest-mac.dmg. This link downloaded version 10.4.  Replacing it in the Installomator script successfully downloaded and installed the application.

naschenbrenner:
crashplan split off from code42 a few months back, they're separate products now so that link unfortunately doesn't point to the right installer called out in this issue.

orrison-CrashPlan:
Hi all, I'm an employee over at CrashPlan weighing in to try and help disambiguate things just a tad. 

We at CrashPlan did split off from Code42 a few months ago. We also just released our version 11 app which is branded CrashPlan and so should make things a bit clearer moving forward. At this point we should have three different labels for CrashPlan/Code42. I can't fully comment on the Code42 piece since I am not employed by Code42.

1. A Code42 label, the one currently in place I suspect will work great. 
https://github.com/Installomator/Installomator/blob/main/fragments/labels/code42.sh

2. A CrashPlan label. Built like this:

```
crashplan)
    name="CrashPlan"
    type="pkgInDmg"
    downloadURL="https://download.crashplan.com/installs/agent/latest-mac.dmg"
    expectedTeamID="UGHXR79U6M"
    blockingProcesses=( NONE )
    ;;
```
3. A CrashPlan small business label as discussed in this pull request. That will have a similar download URL, and team ID as the main CrashPlan one.
https://github.com/Installomator/Installomator/pull/931

lennyvaknine43:
[...]

PR #766, opened 2022-11-15
updated code42 download url's with new hostname 

naschenbrenner:
Code42 adjusted their download hostname to be https://download-preservation.code42.com making the old hostname in the "downloadurl" no longer valid. This will still pull down an older version as noted in #572 but will allow the workflow to still function for organizations with the option to upgrade via the cloud.


scriptingosx:
Thank you!

PR #778, opened 2022-11-22
Use Self Service icon as the 'overlayicon'

dan-snelson:
Thanks for all your work on `v10`.

This PR uses the Self Service icon as the 'overlayicon' in the `mini` progress dialog.

Screenshot 2022-11-22 at 9 36 31 AM


Theile:
Thank you, it looks great. It's only for Jamf, so think we need to consider if it's running in other MDM. We have tests for that, so maybe a test for that here as well.

I can help, I'm just thinking loud here.

Theile:
Oh, I realise you only changed the supporting files for Jamf. I will rest my rambling of other MDM solution, but probably copy your idea (again, again).

Thank you, I put it on the milestone.

dan-snelson:
Thanks, @Theile.

scriptingosx:
Thank you!

PR #783, opened 2022-11-23
Adding processing for new label property 'folderName', adds chemdoodle label

bartreardon:
This update allows for cases where the app is within a folder within a DMG where the distribution requires the parent folder containing the app needs to be copied to the destination, not just the app itself (should also work for ZIP but I haven't come across an example where this would be required).

Example:
`AppName.dmg` when mounted contains a folder `AppFolderName` which contains the app bundle but also other related content (companion apps, settings etc)
When mounted the folder structure will be `/Volumes/AppName/AppFolderName/AppName.app`
This change will let you specify `AppFolderName` in the label. Verification will still occur on `AppName.app` after which `AppFolderName` and its contents will be copied into place.

Changes are: 
 - New label property `folderName`
 - Update to `installAppWithPath` to take a `folderPath` argument. Will still perform all the same checks on `$appPath` but instead of `ditto -v "$appPath"` to install the app it performs `ditto -v "$folderPath"` 
 - Function `installFromDMG()` updated to handle `folderPath`
 
As written, `installFromDMG` will always pass in two arguments to `installAppWithPath`. The check in `installAppWithPath` will verify if it has an actual value or `"$dmgmount/` with no folder path by looking for a trailing `/`. Optionally could make `installFromDMG` only pass the argument if required but we are doing the verification anyway.

Also, this change will preserve removing of the app bundle if an earlier version exists, but does not remove the parent folder. This is partially intentional as there may be app configuration however it may be desirable to either allow (or explicitly disallow) removal of the parent folder during an update. In testing the current behaviour didn't present an issue though.
 
Included in the PR is a label for an app where this is the case (ChemDoodle) but the format would look something like:

```bash
applabel)
     name="AppName"
     type="dmg"
     downloadURL="https://some/url/AppName.dmg"
     expectedTeamID="ABC1234"
     folderName="AppFolderName"
     appName="${folderName}/AppName.app"
     appNewVersion=$(...)
     ;;
```
 
Tested with ChemDoodle and also verified this should not break any existing functionality (not in the labels I tested at least)

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -4144,6 +4144,18 @@ jetbrainswebstorm)
     appNewVersion=$( curl -fsIL "${downloadURL}" | grep -i "location" | tail -1 | sed -E 's/.*\/[a-zA-Z-]*-([0-9.]*).*[-.].*dmg/\1/g' )
     expectedTeamID="2ZEFAR8TH3"
     ;;
+jre8)
+    name="Java Runtime Environment 8"
+    type="pkgInDmg"
+    versionKey="CFBundleVersion"
+    versionURL=$(curl -fs "https://javadl-esd-secure.oracle.com/update/mac/map-mac-1.8.0.xml" | xpath '( //java-update-map/mapping/url)[last()]' 2>/dev/null | cut -d\> -f2 | cut -d\< -f1)
+    appNewVersion=$(curl -fs "${versionURL}" | xpath '(//rss/channel/item/enclosure/@sparkle:version)' 2>/dev/null | cut -d '"' -f 2)
+    appBuildVersion=$(echo $appNewVersion | cut -d. -f3)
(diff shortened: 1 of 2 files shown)


bartreardon:
oh - it added the last commit to this pr....was hoping to make that a seperate pr. :/

Theile:
Wow, these last changes are for me. Correcting `--moveable`. I'm working on adding `--overlayicon` as well. You can revert the commit if you like. I got this. Thanks!

bartreardon:
Can do 🙂 - although `movable` is a perfectly cromulent word so I'm going to add that in as a valid spelling of the argument.

bartreardon:
Found a bug in the chemdoodle labels introduced as the vendor has updated their download page somewhat. the downloadURL should work correctly now.

bartreardon:
I need to re-submit this PR - too many files got added to other branches and it was a pain to try and keep things in check.

bartreardon:
oh wait - I don't need to as that is included here. didn't realise it was merged in https://github.com/Installomator/Installomator/commit/ed83af9c8fc655d67e74066c13183ced38b596f7

PR #793, opened 2022-12-02
Create lsagent.sh

Theile:
Lansweeper agent (LsAgent).


Theile:
I have now added internal variables when this label is called. That makes it possible to install this fully automated. I have only tested with `server`, `port`, and `mode`. I expect that if `agentkey` is used, then `server` and `port` is not.

scriptingosx:
I like the idea, but cannot verify that it works without having valid values?

PR #794, opened 2022-12-02
When available, include "Ver. `$appNewVersion`" in `progresstext`

dan-snelson:
Screenshot 2022-12-02 at 7 26 57 AM
Screenshot 2022-12-02 at 7 27 07 AM


dan-snelson:
I think I prefer `v` …

Screenshot 2022-12-02 at 9 14 11 AM
Screenshot 2022-12-02 at 9 14 20 AM

… also, introduces `appVersionProgressText` so that `appNewVersion` isn't modified in case it's used later on.

PR #797, opened 2022-12-05
Update Alfred label

alexfinn:
Alfred 4.app is now Alfred 5.app, Label is tested and working nicely with a simple edit of the number 4 to 5


scriptingosx:
Would people have reason to stick with Alfred 4? Should there be a new label for legacy Alfred 4?

scriptingosx:
please fix this pr to change the label in `fragments/labels/alfred.sh` not the in the main script

PR #803, opened 2022-12-09
Fix for Microsoft Teams version

Theile:
Fixed locating the latest version and enabled `updateTool`. 

NOTE that `updateTool` requires Terminal and the Jamf binary to be able to send Apple events to Microsoft AutoUpdate. PPPC profile from here can do that:

https://github.com/pbowden-msft/MobileConfigs/tree/master/Jamf-MSUpdate

### DIFF
--- a/fragments/labels/microsoftteams.sh
+++ b/fragments/labels/microsoftteams.sh
@@ -1,13 +1,17 @@
 microsoftteams)
     name="Microsoft Teams"
     type="pkg"
-    packageID="com.microsoft.teams"
+    #packageID="com.microsoft.teams"
     downloadURL="https://go.microsoft.com/fwlink/?linkid=869428"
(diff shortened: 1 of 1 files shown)


scriptingosx:
What happens when the profile is not there? I.e. what error occurs then? can we add that to a comment in the label?

Theile:
@isaacatmann can you disclose details about this?

Theile:
"msupdate works by using native XPC to communicate with the MAU daemon. On macOS 10.14 Mojave and later, you may see a privacy prompt when running msupdate for the first time. If you're using an enterprise management tool such as Jamf Pro, you should deploy a Privacy Preferences Policy Control (PPPC) payload to pre-approve access. Samples of such a payload can be downloaded from GitHub."

https://learn.microsoft.com/en-us/deployoffice/mac/update-office-for-mac-using-msupdate

https://github.com/pbowden-msft/MobileConfigs/tree/master/Jamf-MSUpdate

scriptingosx:
Thank you!

PR #826, opened 2022-12-29
Create microsoftonedrive-rollingout.sh

Theile:
Seems like Jamf Pro is using a later version than the current released latest version in their Patch management system. OneDrive app itself does not know about this version, and it has a different download URL. 

The version Jamf Pro uses is "Rolling out" from this web page: https://support.microsoft.com/en-us/office/onedrive-release-notes-845dcf18-f921-435e-bf28-4e24b95e5fc0

Here the download URL is: https://go.microsoft.com/fwlink/?linkid=861011

### DIFF
--- a/fragments/labels/microsoftonedrive-rollingout.sh
+++ b/fragments/labels/microsoftonedrive-rollingout.sh
@@ -1,0 +1,14 @@
+microsoftonedrive-rollingout)
+    name="OneDrive"
+    type="pkg"
+    downloadURL="https://go.microsoft.com/fwlink/?linkid=861011"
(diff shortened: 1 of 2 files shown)


scriptingosx:
I don't understand the modifications to the original label?

Theile:
Idea is that if it's called from Jamf Pro (then it should change to the rollout-version).

I don't know how often this problem will persist, or if it's important enough to solve like this. I don't know why Jamf Pro would use the Rollout version number and not the current version in the first place.

Maybe we should let the standard label be as it was, and just add the rollout one. But not sure if Jamf is persistent on showing the rollout version always. I have personally just changed the workflow.

Theile:
I removed the changes to the original label. So now we can choose either one. Currently it's a problem again with the version in Jamf Pro.

scriptingosx:
Thank you!

PR #833, opened 2023-01-06
Use swiftDialog for notifications (if installed)

Theile:
If `/usr/local/bin/dialog` exist then we can use that for notifications

Could be neat to use `icon=` to show the icon if the app that the notification is about, but currently we have no variable for that.


Theile:
A full install of VLC…
Screenshot 2023-01-06 at 11 29 28

PR #847, opened 2023-01-12
Fix Loom download URL

pythoninthegrass:
Upstream YAML file changed. Negative match for arm64 on i386 arch. Piping to head was unnecessary.

Spoke with @acodega on the [MacAdmins Slack](https://macadmins.slack.com/archives/C013HFTFQ13/p1673483194062239) regarding the fix.


scriptingosx:
Thank you!

PR #854, opened 2023-01-21
Update code for Google Ads Editor (googleadseditor)

gilburns:
Updated code for version checking of Google Ads Editor. This will check online version and local version. Slightly hackish, but works for now. :-)


scriptingosx:
Thank you!

PR #860, opened 2023-01-23
appClosed being set prematurely

danwitkowski:
The variable `appClosed` is being set to `1` if a blocking process is found, but before any processes are actually closed. This has unintended behavior in the case where `BLOCKING_PROCESS_ACTION` is `silent_fail` since nothing is actually closed in this case statement, only the `cleanupAndExit` function is called.

This function calls `reopenClosedProcess` which checks whether `appClosed` is set and tries to reopen the app if so. Because the app was already open (which is why we entered this case statement in the first place) and the variable was set prematurely, all this does is instead pop the app into focus on the user's screen, interrupting whatever it was they were doing.

This was noticed due to any app that we attempted to update with this method popping into focus one or more times a day.

I added an additional function to set this value in cases where Apple Script is being used to try and closed the app, so that this variable gets set correctly in each case. I also set the variable for the case where instead of Apple Script, the `pkill` command is used.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -635,6 +635,12 @@ getAppVersion() {
     fi
 }
 
+quitApp() { # $1 = app name
+    printlog "telling app $ to quit"
+    runAsUser osascript -e "tell app \"$x\" to quit"
+    appClosed=1
+}
+
 checkRunningProcesses() {
     # don't check in DEBUG mode 1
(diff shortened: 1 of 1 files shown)


scriptingosx:
duplicate of #657 which is already merged into the `dev` branch for v11

PR #864, opened 2023-01-24
Proxy Server Connectivity Optional Parameters

dionparsons:
Optional Setting for Proxy to be configured, JAMF Parameter Options 7 = Proxy Server and Parameter Options 8 = Proxy Port
If Parameters not set then blank values are put in for HTTP and HTTPS Proxy

### DIFF
--- a/MDM/Jamf/00_PrepareInstall_SwiftDialog.sh
+++ b/MDM/Jamf/00_PrepareInstall_SwiftDialog.sh
@@ -15,6 +15,22 @@ icon=${6:-"/System/Applications/App Store.app/Contents/Resources/AppIcon.icns"}
 # see Dan Snelson's advice on how to get a URL to an icon in Self Service
 # https://rumble.com/v119x6y-harvesting-self-service-icons.html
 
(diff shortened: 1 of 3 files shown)


dionparsons:
@acodega Spelling Errors Fixed and Variables formatted with All Caps

scriptingosx:
Installomator does not use fixed arguments like $7, $8 etc. doing that would break the other methods of supplying arguments.

Is there a way to pass the proxy config into `curl` with arguments? because then we should be able to leverage `curlOptions` without needing to change the code?

dionparsons:
Using Curl options would mean having to modify all labels with a Curl command wouldn’t it?

I’ll look to modify the variable input method, see if that helps.

Ideally the proxy only gets used when Installomator.sh session is used and doesn’t affect the system, pretty sure it’s only a session proxy and thus applies to all outbound traffic from script not just Curl commands.

scriptingosx:
> Using Curl options would mean having to modify all labels with a Curl command wouldn’t it?

Good point... There are ways. We could overload `curl` with a function of the same name (we do that with `xpath`) but that might break other uses.

dionparsons:
Rewriting Code

PR #870, opened 2023-01-27
Proxy Patch

dionparsons:
I have added two variables PROXYPORT and PROXYSERVER.

If these values are set then a ping test will be performed, if ping is success then proxy will be set as per variables.

Setting the proxy is inserted after JAMF Parameter 1380 Parsing Parameters in Arguments.sh and Shifting so that you can set the following values in JAMF and it function correctly. 

The Variables are declared in the Header.sh

Parameter 6 - PROXYSERVER=192.168.10.20
Parameter 7 - PROXYPORT=8080

Printlog has been added to provide more information of when and if the script is applying the proxy settings.

### DIFF
--- a/fragments/arguments.sh
+++ b/fragments/arguments.sh
@@ -110,6 +110,22 @@ if [[ ! -x $DIALOG_CMD ]]; then
     DIALOG_CMD_FILE=""
 fi
 
+# check for proxy and set
+if [[ -z $PROXYSERVER ]];then
+	printlog "Proxy IP Address not Set, Skipping Set Proxy" INFO
+else
+    printlog "Proxy value set to $PROXYSERVER:$PROXYPORT" INFO
+	if ping -q -c 1 -W 5 $PROXYSERVER; then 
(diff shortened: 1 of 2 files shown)


acodega:
Does cURL automatically use those values if they are set? I'm not following.

acodega:
Oh, wow. https://oxylabs.io/blog/curl-with-proxy

dionparsons:
Setting the Proxy in the Session, Curl then uses the Session Proxy. Using Proxy in Environment Variables.

dionparsons:
1. You could put both server name and port in one variable reducing the amount of variables

2. Some programs or legacy systems may have used uppercase proxy so usually you set lower case and upper case see https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca

Let me know if you want an amendment it looks as though the code now conflicts.

acodega:
@dionparsons Can this PR be modified so we can merge it? It'd be very useful to have this feature.

PR #874, opened 2023-01-28
Update jdk18.sh

macmangr:
Label adjusted because Java SE 18 was removed from the "Latest" area and moved to the "Archive"


scriptingosx:
Thank you!

PR #883, opened 2023-02-06
Update musescore.sh

sintichn:
musescore released version 4, the name musescore 3 produces an exit code of 8. Path not found.


scriptingosx:
Thank you!

PR #885, opened 2023-02-07
Update microsoftoffice365.sh

ricardtolosa:
Modifying the definition of "appNewVersion" as the new installer name changed and it wasn't returning the version number.

New PKG name: Microsoft_365_and_Office_16.69.23010700_Installer.pkg


scriptingosx:
Thank you!

Issue #888, opened 2023-02-08
Draw.io is installing the Apple Silicon build in stead of the Universal Build


timnottom:
The draw.io label should be updated to install the Universal build.

timnottom:
See: https://github.com/Installomator/Installomator/pull/889
Would be nice if someone would doublecheck this.

PR #889, opened 2023-02-08
drawio: Added archiveName to download the Universal build

timnottom:
Without the addition of archiveName the ARM64 build was downloaded.


timnottom:
See: https://github.com/Installomator/Installomator/issues/888

scriptingosx:
Thank you!

PR #899, opened 2023-02-16
Add ignore-update-tool option to the INSTALL parameter

Maarten-Wijnants:
This option allows you to ignore the update tool when it is defined in a label like googlechromepkg.

It can be useful if you want to make sure updates are applied and don't want to depend on the end user restarting the application.


Theile:
I don't think we need this PR. We can already disable updateTool by sending `updateTool=""` when calling Installomator.

PR #907, opened 2023-02-22
Update versionFromGit function

wakco:
As not every developer adheres to version numbers containing only numbers and dots, the sed filter should be optional. However as there may be labels that require the sed filter already, this provides the option to ask the function to skip the sed filter.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -513,12 +513,17 @@ downloadURLFromGit() { # $1 git user name, $2 git repo name
 
 versionFromGit() {
     # credit: Søren Theilgaard (@theilgaard)
-    # $1 git user name, $2 git repo name
+    # $1 git user name, $2 git repo name, $3 filtered/unfiltered (optional, default filtered (limits version to numbers and dots only))
     gitusername=${1?:"no git user name"}
     gitreponame=${2?:"no git repo name"}
-
+    gitfiltered=${3?:"filtered"}
+    
     #appNewVersion=$(curl -L --silent --fail "https://api.github.com/repos/$gitusername/$gitreponame/releases/latest" | grep tag_name | cut -d '"' -f 4 | sed 's/[^0-9\.]//g')
-    appNewVersion=$(curl -sLI "https://github.com/$gitusername/$gitreponame/releases/latest" | grep -i "^location" | tr "/" "\n" | tail -1 | sed 's/[^0-9\.]//g')
+    #appNewVersion=$(curl -sLI "https://github.com/$gitusername/$gitreponame/releases/latest" | grep -i "^location" | tr "/" "\n" | tail -1 | sed 's/[^0-9\.]//g')
+    appNewVersion=$(curl -sLI "https://github.com/$gitusername/$gitreponame/releases/latest" | grep -i "^location" | tr "/" "\n" | tail -1)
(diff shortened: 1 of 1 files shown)


PicoMitchell:
@scriptingosx This line is also problematic for versions which don't only contain dots an numbers: https://github.com/Installomator/Installomator/blob/b46a0f3437594a0488e85e7c8600749e59b684cc/fragments/functions.sh#L240

For example:
```
pkgutil --pkg-info-plist org.freegeek.pkg.mkuser 2>/dev/null | grep -A 1 pkg-version | tail -1 | sed -E 's/.*>([0-9.]*)<.*/\1/g' 
```

Incorrectly outputs the following entire plist line because the `sed` expression didn't match the version string.:
```
	<string>2022.9.30-1</string>
```

That line could be replaced with the following to reliably extract any version string from the plist regardless of the characters it contains without needing the `grep | tail | sed`:
```
/usr/libexec/PlistBuddy -c 'Print :pkg-version' /dev/stdin <<< "$(pkgutil --pkg-info-plist "${packageID}" 2>/dev/null)" 2> /dev/null
```

Retrieving the latest version from GitHub could also be done like this (or the `getJSONValue` function that is already in `installomator` could be used instead of having a standalone JXA command here):
```
osascript -l 'JavaScript' -e 'run = argv => JSON.parse(argv[0]).tag_name' -- "$(curl -sfL "https://api.github.com/repos/$gitusername/$gitreponame/releases/latest")" 2> /dev/null
```

I think I would also argue that there is no `unfiltered` option needed at all when the actual raw version strings are retrieved properly no matter what they are.

scriptingosx:
@PicoMitchell those comments are fair and appreciated, but can you open a new issue or create a PR where they are fixed so we can track these separately? unless @wakco wants to add to this PR?

wakco:
This PR was about introducing a workaround for a bug I reported, because my report was mis-understood. @PicoMitchell has provided a much better explanation of the bug, so if the bug gets fixed I'm happy either way, because it should mean the workaround is no longer required.

In short since best information on the bug is here, I'm happy for it to be added and tracked here, however I'm not sure I am the best person to do that work.

PicoMitchell:
I'll try to do some PRs this weekend for this and other little things I've noticed.

In regards to version retrieval from GitHub, it looks like it was an intentional change to move away from using the API URL in the past. Was the API not working properly in some cases? I wouldn't want to submit a PR to bring back something that was already decided to move away from for good reasons.

It looks like @Theile made the change here https://github.com/Installomator/Installomator/commit/1de3838a9b1dff13f0dab4f84eb19313670f3e48, so pinging him since maybe he remembers why that change was made.

Also, upon further consideration I'm assuming that some version number filtering from GitHub tags is actually required for some apps since those tags can be more than just the raw version number. (But, versions from `pkgutil` should likely never be altered since those would always be actual version numbers, same as versions from `Info.plist` files are not altered.)

PicoMitchell:
I've just submitted https://github.com/Installomator/Installomator/pull/914 to  resolve extracting the package version from `pkgutil --pkg-info-plist`.

I think this PR could be accepted as is to resolve the immediate issues with retrieving versions from GitHub using the tag names.

Whether or not the GitHub API URLs should be used for the tag name retrieval would be a separate discussion.

PR #912, opened 2023-02-26
Add "valueFromJSON" to eventually replace "getJSONValue"

PicoMitchell:
This `valueFromJSON` function is my latest `json_value` (Version 2023.2.26-1) function (https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh) but renamed to match the Installomator function naming style.

It is intentionally NOT yet replacing the existing `getJSONValue` function because both need to exist simultaneously until all labels have been updated to use the new `valueFromJSON` function. I plan to work through those and submit PRs for each label that is currently using `getJSONValue`, and then will submit a PR to remove `getJSONValue` when Installomator and all labels have been updated and merged. I think giving this function a new name is graceful way to handle this transition while also still having a function name that is intuitive and fits in with other functions in Installomator.

This new `valueFromJSON` function has a handful of benefits over the existing `getJSONValue` function and the only drawback is it's a bit more code, and retrieving nested values is not backwards compatible with the current `getJSONValue` function.

Some of the benefits are of `valueFromJSON` are:

- Can accept stdin so that is can be used like `curl ... | valueFromJSON ...` which can make the code in some labels cleaner and simpler.

- Can accept JSON as the first or last argument so it's flexible for different users styles in labels.

- Does not parse any key or index arguments passed to it directly has JXA code like the current `getJSONValue` function does. This means advanced JavaScript methods cannot be used in `valueFromJSON`, but that is not currently being taken advantage of in Installomator and should not be needed for basic version or URL retrievals from JSON, and the way `valueFromJSON` handles arguments is safer overall.

- No knowledge of JavaScript dot or bracket notation is required to use `valueFromJSON` like it is for `getJSONValue` to retrieve nested values or complex key name. To retrieve a value value from a dictionary that is nested in an array with `getJSONValue` the expression `getJSONValue '[0].keyName'` needs to be passed as a single argument. But, with `valueFromJSON` each index or key is simply passed as its own argument like `valueFromJSON '0' 'keyName'`. This also means there are no possible characters that could exist in valid key names that would need to be escaped or placed in bracket notation to properly retrieve a key. One example would be a period in a key would need to be retrieved with bracket notation like `getJSONValue '["key.name"]'` which could be confusing for some users when `getJSONValue 'key.name'` does not work as expected (this would also apply for any possible key names that contain spaces). With `valueFromJSON`, using `valueFromJSON 'key.name'` would properly retrieve the key with the literal name `key.name`.

- Negative array indexes can specified be used when someone needs to retrieve the last value of an array.

- `valueFromJSON` has much better error handling and would output an error to stderr with the name of the key or index that was not retrieved rather than outputting nothing like `getJSONValue` would in some cases.

- Detailed help information is documented at https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh and this URL is commented in the function which may be helpful for users who haven't worked with JSON parsing tools before.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -156,7 +156,7 @@ valueFromJSON() { # Version 2023.2.26-1 - Copyright (c) 2023 Pico Mitchell - MIT
         -e 'parse(json); (i === 0 ? argv.shift() : (i === 1 && argv.pop())); break } catch (e) {} } if (out === undefined) throw "Failed to parse JSON."; argv.forEach(key => {' \
         -e 'out = (Array.isArray(out) ? (/^-?\d+$/.test(key) ? (key = +key, out[key < 0 ? (out.length + key) : key]) : (key === "=" ? out.length : undefined)) : (out instanceof' \
         -e 'Object ? out[key] : undefined)); if (out === undefined) throw "Failed to retrieve key/index: " + key }); return (out instanceof Object ? JSON.stringify(out, null, 2)' \
-        -e ': out) }' -- "$@" 2>&1 >&3)"; } 3>&1; [ "${1##* }" != '(-2700)' ] || { set -- "json_value ERROR${1#*Error}"; >&2 printf '%s\n' "${1% *}"; false; }
+        -e ': out) }' -- "$@" 2>&1 >&3)"; } 3>&1; [ "${1##* }" != '(-2700)' ] || { set -- "valueFromJSON ERROR${1#*Error}"; >&2 printf '%s\n' "${1% *}"; false; }
 }
 
 # will get the latest release download from a github repo
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -148,6 +148,17 @@ deduplicatelogs() {
     done <<< "$loginput"
 }
 
+valueFromJSON() { # Version 2023.2.26-1 - Copyright (c) 2023 Pico Mitchell - MIT License - Full license and help info at https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh
+    { set -- "$(/usr/bin/osascript -l 'JavaScript' -e 'function run(argv) { const stdin = $.NSFileHandle.fileHandleWithStandardInput; let out; for (let i = 0; i < 3; i ++) {' \
+        -e 'let json = (i === 0 ? argv[0] : (i === 1 ? argv[argv.length - 1] : $.NSString.alloc.initWithDataEncoding((stdin.respondsToSelector("readDataToEndOfFileAndReturnError:")' \
+        -e '? stdin.readDataToEndOfFileAndReturnError(ObjC.wrap()) : stdin.readDataToEndOfFile), $.NSUTF8StringEncoding).js.replace(/\n$/, ""))); if ($.NSFileManager.defaultManager' \
+        -e '.fileExistsAtPath(json)) json = $.NSString.stringWithContentsOfFileEncodingError(json, $.NSUTF8StringEncoding, ObjC.wrap()).js; if (/[{[]/.test(json)) try { out = JSON.' \
+        -e 'parse(json); (i === 0 ? argv.shift() : (i === 1 && argv.pop())); break } catch (e) {} } if (out === undefined) throw "Failed to parse JSON."; argv.forEach(key => {' \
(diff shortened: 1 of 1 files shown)


PicoMitchell:
Maybe the function name `getValueFromJSON` fits better. I'm open to calling it whatever you all think fits best.

scriptingosx:
I appreciate the effort, but is there some way we can achieve improvement of the existing function without redundancy?

PicoMitchell:
The redundancy will be removed after I update all the labels that are currently using `getJSONValue`, which is only like 15 or so I think.

scriptingosx:
ok, let me rephrase: is there some way we can improve the existing function without needing to replace it entirely?

Your json_value function is powerful but quite complex and we have concerns that we will need to maintain it going forward.

PicoMitchell:
I can't imagine what would break or need maintenance in what it's doing. Everything it's doing it quite simple. If you read through the raw JavaScript source with formatting, I think you'll agree it's all pretty basic stuff (not like the `json_extract` function which is more complex): https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/JXA%20Source%20Files/json_value.jxa

scriptingosx:
we can _never_ imagine what would break or need maintenance beforehand... Given those unknowns, _less_ is generally _less_ work to maintain

PicoMitchell:
Yeah, but any future changes in JavaScript versions is not going to break backwards compatibility. Anything that might break in what this is doing is just as likely to break the current `getJSONValue` function.

In regards to Apple API breaking, I'm only using fundamental stuff, both of which are also used in the current `getJSONValue` function, so that's no different a vulnerability than currently exists (which I don't think is really a vulnerability at all).

If code simplicity was the goal, this one-liner version or any of the other simpler variants could be used and all would be safer than the current getJSONValue function, just without as many convenience features as the currently suggested variant has such as file reading (which is not used by Installomator), reading stdin (which would be nice to use but could be lived without), negative array indexes (which probably would come in handy but also not critical): https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh#L147-L149

This one-liner variant uses absolutely basic JavaScript functionality and zero native API. Also the variant which adds negative index handling for arrays is still quite basic JavaScript functionality and also doesn't use any native APIs either.

PicoMitchell:
If simplicity and code clarity is paramount over line count or features, here is the variant that has negative array indexes and array counts, nice error handing, and JSON output for object values but no file reading, no stdin reading, and JSON input only as the last argument (but that is easy to switch to be only the first argument is that is preferred) but with all the nice whitespace formatting brought back so that it's easy to see what all it's doing:

```
json_value() { # Version 2023.2.26-1 - Copyright (c) 2023 Pico Mitchell - MIT License - Full license and help info at https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh
	{
		set -- "$(/usr/bin/osascript -l 'JavaScript' -e '
function run(argv) {
	let out = argv.pop()
	if (/[{[]/.test(out)) out = JSON.parse(out)

	argv.forEach(key => {
		out = (Array.isArray(out) ?
			(/^-?\d+$/.test(key) ?
				(key = +key, out[key < 0 ? (out.length + key) : key]) :
				(key === "=" ? out.length : undefined)
			) :
			(out instanceof Object ?
				out[key] :
				undefined
			)
		)

		if (out === undefined) throw "Failed to retrieve key/index: " + key
	})

	return (out instanceof Object ? JSON.stringify(out, null, 2) : out)
}
' -- "$@" 2>&1 >&3)"
	} 3>&1

	[ "${1##* }" != '(-2700)' ] || {
		set -- "json_value ERROR${1#*Error}"; >&2 printf '%s\n' "${1% *}"
		false
	}
}
```

(The funny stuff I'm doing with `set` is done to be able to not set any global variables to capture any error output (which is what all the fd redirect stuff is for) within this function since POSIX shell does not have variables that can be set to be `local` to a function like `bash` and `zsh`, and I wrote this to be compatible with all shells. But since Installomator is `zsh`, it could be modified to just use a regular `local` variable for the sake of code clarity if needed.)

Here is the consolidated version of this variant: https://github.com/RandomApplications/JSON-Shell-Tools-for-macOS/blob/main/json_value.sh#L111-L120

Also, in terms of maintainability, unless I get hit by a bus or whatever I plan to keep these JSON shell tools maintained and tested with all new versions of macOS and updated as needed.

If this variant would be more be preferred to use in Installomator, I can push a new commit to this PR with this function.

PR #916, opened 2023-02-26
Pass variables as args to all "osascript" commands

PicoMitchell:
Instead of placing variables directly in AppleScript code passed to  "osascript", pass the variables as arguments to be retrieved from within the AppleScript code instead.

To be able to retrieve arguments passed to "osascript" within the AppleScript code, the AppleScript code must be within an "run" function/handler, which is shown and described near the top of the "DESCRIPTION" section in "man osascript".

When a variable is placed directly in AppleScript code, its value is expanded by the shell before being passed to "osascript" and is therefore being interpreted as raw AppleScript code. This means that any special characters such as double quotes (") would need to be manually escaped within the variables contents to not break the quoted AppleScript string that the variable is being placed within.

But, when the values are passed to "osascript" as arguments and then retrieved within the AppleScript code from the arguments array, they are always static string values that are never interpreted as AppleScript code. This means that any literal quotes within a variable are properly displayed without needing to be escaped and without ever breaking the AppleScript code.

Double quotes are just one example, but in general passing variables as arguments prevents any possible arbitrary AppleScript code execution. While that is not really likely with how these function are used in Installomator, passing arguments to other environments that interpret code instead of placing variables directly in that code is a safe practice in general and helps avoids any possible mistakes or bugs in the future.

Also, I have slightly adjusted the behavior of "displaydialog" function to set the "Not Now" button as the dialogs "cancel button" which means the Escape key can now be used to dismiss the alert as is normal and expected in standard macOS dialogs. The adjustments needed for this behavior to not require other changes throughout the code is documented in the "NOTE" comment at the end of the "displaydialog" function.

Finally, I added a new "quitApp" function instead of repeating the same "osascript" command 4 times to quit apps in different case statements within the "checkRunningProcesses" function.

These may be a lot of changes in a single PR, but they were all pretty interconnected. If any part of all these changes is not desired, I can make changes as needed.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -53,14 +53,14 @@ displaydialog() { # $1: message $2: title
     title=${2:-"Installomator"}
     #runAsUser osascript -e "button returned of (display dialog \"$message\" with  title \"$title\" buttons {\"Not Now\", \"Quit and Update\"} default button \"Quit and Update\" with icon POSIX file \"$LOGO\")"
     runAsUser osascript -e '
-on run argv
+on run {dialogMessage, dialogTitle, dialogIconPath, dialogTimeoutSeconds as integer}
     try
-        return (button returned of (display dialog (text item 1 of argv) with title (text item 2 of argv) buttons {"Not Now", "Quit and Update"} cancel button 1 default button 2 with icon (POSIX file (text item 3 of argv))))
+        return (button returned of (display dialog dialogMessage with title dialogTitle buttons {"Not Now", "Quit and Update"} cancel button 1 default button 2 with icon (POSIX file dialogIconPath) giving up after dialogTimeoutSeconds))
     on error
         return "Not Now"
     end try
 end run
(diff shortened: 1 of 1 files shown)


PicoMitchell:
Would you like me to re-do this as a PR agains the Dev branch to include the `giving up after $PROMPT_TIMEOUT` option in `displaydialog`?

If so, would I need to close this and re-do a PR with a branch from Dev, or is there an easy way for me to commit an update to this PR that resolves the Dev conflicts?

Theile:
Looks to me like `quitApp $x` could be a little dangerous, as the function only grabs `$1`, so “Microsoft Edge” would only be “Microsoft”. I think it should be `quitApp "$x"`.

(But today I learned something about `osascript`, thank you!

PicoMitchell:
In `zsh` word splitting would not happen on the `$x` argument passed to `quitApp`, so `$1 `would always be the whole thing, such as "Microsoft Edge"

In `bash` or POSIX shell (`sh`), it would be important to quote the variable passed to `quitApp`, like `quitApp "$x"`, but that's not an issue since Installomator is `zsh`.

PicoMitchell:
I generally write in `bash` so I'm still getting the hang of `zsh` quoting and word splitting behavior. Totally trips me up that passing unquoted command substitution as an argument is word split but an unquoted variable is not.

PR #917, opened 2023-02-26
Extract "Distribution" XML to get versions from downloaded packages using "xar" instead of "pkgutil --expand"

PicoMitchell:
All information about the why's and how's of extracting only the "Distribution" XML file from a package using "xar" have been written into the code comments. Basically, extracting ONLY the "Distribution" XML from the package can be a lot more efficient when dealing with large packages rather than extracting the entire package contents which could include large "Payload" files, etc.

But, since "man xar" has a deprecation warning, this code checks if "xar" returned nothing in case of future removal or breakage and will fallback on extracting the entire package contents using "pkgutil --expand" like is currently being done. So, I think this is a win win since either the process is more efficient or the same as it currently is no matter what happens to "xar" in the future.

If no version is detected either because the package is not a "distribution package" and therefore doesn't have a "Distribution" XML file or the XPath expression failed to retrieve a version for some reason, those errors will be logged by using new "appNewVersion" error values which would always just fail the "if [[ $appversion == $appNewVersion ]]; then" condition to always install the downloaded package when a version couldn't be detected for any reason.


PicoMitchell:
I see that https://github.com/Installomator/Installomator/blob/f07254cf4a90287d4890c3475d7f08b1daea5f73/fragments/functions.sh#L664 in the `dev` branch has an improved XPath expression to retrieve the version for the specific `packageID`.

I did not include this change since I was working off the `main` branch, but wanted to point it out depending on if this change get merged into a minor or major release.

PicoMitchell:
I think the `PackageInfo` XML file could also be checked for a version for non-distribution/component style packages.

But I think it's not uncommon for folks to not include `--version` when just using `pkgbuild`. With no version specified in `pkgutil` or `productbuild` it will default to "0". It seems like "1.0" is also a common value so maybe other package building tools set that default value and folks don't update it.

scriptingosx:
The reason we are expanding the entire archive is that the `xar` command did not have these options in older versions of macOS. How far back does this work?

PicoMitchell:
Just confirmed `xar` and all the options I'm using are available at least back to High Sierra, and also double-check Mojave. But, with how I wrote the fallback, if `xar` or at least the `-t` option did end up being unavailable on older versions of macOS, `pkgutil --extract` would get fallen back on anyways.

hcbraun:
fyi, I had to update the xpath again to make it work for the Adobe Acrobat package: [xpath patch](https://github.com/Installomator/Installomator/pull/927/files#r1122729114)

PR #922, opened 2023-02-28
Update lsagent.sh

wakco:
agentkey and server options are not mutually exclusive, at my workplace we set both options.


wakco:
I should have died the error message while I was at it (lsagentKey not lsagentCloudKey, and it is at least one of, not OR).

PicoMitchell:
This could be done without repeating the `CLIArguments` by starting the array with options shared by both modes and then adding the different options to an array in separate `if` blocks like this so either or both options get added as needed:

```
CLIArguments=( --mode $lsagentMode --installer-language $lsagentLanguage )

if [[ -n $lsagentServer ]]; then
    CLIArguments+=( --server $lsagentServer --port $lsagentPort )
fi

if [[ -n $lsagentKey ]]; then
    CLIArguments+=( --agentkey $lsagentKey )
fi
```

wakco:
I'll update and re-submit.

PR #924, opened 2023-02-28
Label update lansweeper agent options for server and agentkey can be used at the same time

wakco:
Lansweeper agent options for server and agentkey can be used at the same time, and cleaned up some of the formatting.

Also thanks @PicoMitchell (sorry referenced the wrong Pico in the actual commit), for cleaning up the building of CLIArguments. Using brackets to define a string like that is still new to me, so I had initially followed the format of the original label.


PicoMitchell:
FYI, you can just commit new changes to your branch of an existing PR, no need to close and open a new one in the future.

PicoMitchell:
You could still make these changes as a new commit to your branch this PR: https://github.com/Installomator/Installomator/pull/922#issuecomment-1447586678

wakco:
forgot to take the opportunity to fix the error message.

PR #925, opened 2023-02-28
Create label for panopto

wakco:
Panopto works best with a client that appropriately matches the server version, hence the requirement for the server address.

The initial download URL will show the server version, but it actually re-directs to another link with the clients CFBundleShortVersionString.

### DIFF
--- a/fragments/labels/panopto.sh
+++ b/fragments/labels/panopto.sh
@@ -1,0 +1,11 @@
+panopto)
+    name="Panopto"
+    type="pkg"
+    if [[ -z $panoptoServer ]]; then
+        cleanupAndExit 89 "This label requires more parameters: panoptoServer=hostname, this will ensure the downloaded version is both current and supports your server." ERROR
+    fi
+    downloadURL="http://$panoptoServer/$(curl -Ls http://$panoptoServer/ | grep cacheRoot | cut -d "'" -f 2 | awk -F '\\' '{ printf  $4 $5 $6 }' | awk -F 'x2f' '{ printf $2 "/" $3 "/" $4 }')/Software/Panopto%20Recorder.pkg?arch=None&useCustomBinary=True"
+    # while there is version information in the URL, it rarely directly matches the client that gets downloaded.
+    # need to create a service account createAccount -username panopto_upload -realname panopto_upload -home /var/panopto -shell /sbin/nologin -hiddenUser
+    expectedTeamID="G7FR783UED"
+    ;;
--- a/fragments/labels/panopto.sh
+++ b/fragments/labels/panopto.sh
@@ -12,7 +12,7 @@ panopto)
         dscl . create /Users/panopto_upload dsAttrTypeNative:IsHidden 1
     fi
     downloadURL="http://$panoptoServer/$(curl -Ls http://$panoptoServer/ | grep cacheRoot | cut -d "'" -f 2 | awk -F '\\' '{ printf  $4 $5 $6 }' | awk -F 'x2f' '{ printf $2 "/" $3 "/" $4 }')/Software/Panopto%20Recorder.pkg?arch=None&useCustomBinary=True"
-    packageID="com.panopto.mac"
+    # not setting packageID as the version will never match (i.e. 13.0.0.12345), while appNewVersion matches CFBundleShortVersionString (i.e. 13.0.0)
     appNewVersion=$(curl -s "$downloadURL" | grep "Object moved to" | sed -e 's/.*Version\=\mac\%2f\(.*\)\&amp\;server.*/\1/')
     expectedTeamID="G7FR783UED"
     ;;
--- a/fragments/labels/panopto.sh
+++ b/fragments/labels/panopto.sh
@@ -12,6 +12,7 @@ panopto)
         dscl . create /Users/panopto_upload dsAttrTypeNative:IsHidden 1
     fi
     downloadURL="http://$panoptoServer/$(curl -Ls http://$panoptoServer/ | grep cacheRoot | cut -d "'" -f 2 | awk -F '\\' '{ printf  $4 $5 $6 }' | awk -F 'x2f' '{ printf $2 "/" $3 "/" $4 }')/Software/Panopto%20Recorder.pkg?arch=None&useCustomBinary=True"
-    # While there is version information in the URL, it rarely directly matches the client that gets downloaded (major version only).
+    packageID="com.panopto.mac"
+    appNewVersion=$(curl -s "$downloadURL" | grep "Object moved to" | sed -e 's/.*Version\=\mac\%2f\(.*\)\&amp\;server.*/\1/')
     expectedTeamID="G7FR783UED"
     ;;
--- a/fragments/labels/panopto.sh
+++ b/fragments/labels/panopto.sh
@@ -4,31 +4,6 @@ panopto)
     if [ -z $panoptoServer ]; then
         cleanupAndExit 89 "This label requires more parameters: panoptoServer=host.name.panopto.com, this will ensure the downloaded version is both current and supports your server." ERROR
     fi
-    # The installer pkg will try to create this account and fail when it attempts to update the
-    #   NFSHomeDirectory and home settings (or it triggers a request for permission from the user).
-    # However, the installer pkg will detect if the account has already been created, and use it,
-    #   so manually creating it stops the installer from bothering the user or failing.
-    # The following is based on what the package does, but should handle it better.
-    panoptoUser="/Users/panopto_upload" # Including /Users/ path
-    if [ "$(dscl . read $panoptoUser 2>&1 | grep -c eDSRecordNotFound)" -gt 0 ]; then
-        for ((panoptoUID = 401; panoptoUID < 500; panoptoUID++)); do
-            if [ "$(dscl . -search /Users UniqueID $panoptoUID | wc -l)" -eq 0 ]; then
-                break  
-            fi
-        done 
-        dscl . -create $panoptoUser UniqueID $panoptoUID
-        dscl . -append $panoptoUser AuthenticationAuthority ';DisabledTags;SecureToken'
-        dscl . -create $panoptoUser PrimaryGroupID 1
-        dscl . -create $panoptoUser NFSHomeDirectory /private/var/panopto
-        dscl . -create $panoptoUser UserShell /sbin/nologin
-        dscl . -passwd $panoptoUser this_password_is_disabled
-        dscl . -create $panoptoUser dsAttrTypeNative:IsHidden 1
-    # The package does do more, however the above covers what is skipped by mk_hidden_user.sh when
-    #   the account already exists. Installomator may still require Full Disk Access, but should
-    #   inherit that permission from the calling process, which doesn't happen for mk_hidden_user.sh.
-    # mk_hidden_user.sh will fail (or request permission) if the NFSHomeDirectory setting is missing,
-    #   because it will attempt to remove the account first.
-    fi
     downloadURL="http://$panoptoServer/$(curl -Ls http://$panoptoServer/ | grep cacheRoot | cut -d "'" -f 2 | awk -F '\\' '{ printf  $4 $5 $6 }' | awk -F 'x2f' '{ printf $2 "/" $3 "/" $4 }')/Software/Panopto%20Recorder.pkg?arch=None&useCustomBinary=True"
     # not setting packageID as the version will never match (i.e. 13.0.0.12345), while appNewVersion matches CFBundleShortVersionString (i.e. 13.0.0)
     appNewVersion=$(curl -s "$downloadURL" | grep "Object moved to" | sed -e 's/.*Version\=\mac\%2f\(.*\)\&amp\;server.*/\1/')
--- a/fragments/labels/panopto.sh
+++ b/fragments/labels/panopto.sh
@@ -11,20 +11,19 @@ panopto)
     # However, the installer pkg will detect if the account has already been created, and use it,
     #   so manually creating it stops the installer from bothering the user or failing.
     # The following is based on what the package does, but should handle it better.
-    panoptoUser="panopto_upload"
-    panoptoPath="/Users/$panoptoUser"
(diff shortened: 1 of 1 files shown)


PicoMitchell:
Creating a user account within a label feels potentially problematic since if somehow this label is run during initial setup before any other user has been created, this user will "steal" the first Secure Token on macOS 11 Big Sur and newer, even if no password is being set since `sysadminctl -addUser` is actually setting a blank/empty password, this account would also take a 50X UID which may not be desirable.

It would be possible to create a "Service Account" style user which could be prevented from being granted the first ST and could also not have any password set at all rather than a blank/empty password, but that would need to be done with a handful of `dscl` commands.

While it isn't uncommon for packages to create "Service Account" users, having installomator take on that role and responsibility feels like it may be out-of-scope, but that would have to some someone's call whose responsible for what Installomator handles.

PicoMitchell:
The bigger question is maybe why/how the package is failing to create the necessary user?

wakco:
macOS since Mojave, Catalina, of Big Sur has stopped allowing installer packages to add users without user approval. There is an MDM config. profile method available to allow installer to add users, but that would then allow any installer package to add users, I recommend manually installing instead. As for the UID and service account, they do not affect Panopto in anyway, and I believe there are other rules from Apple they may imply there are issues with the account the Panopto wants created.

As for the timing of when the label is executed, we could add something this to the label as the first check:
```
if [ "$(who | grep console | cut -d ' ' -f 1) = "_mbsetupuser" ]; then
 cleanupAndExit 89 "Error: Panopto requires the addition of the panopto_upload account, however installing during PreStage Enrolment into an MDM will break the enrolment."
fi

```
Yes there are other ways to detect a logged in user, but the who command is the only one that will advertise _mbsetupuser during a PreStage enrolment at the time the computer will be performing a DEP/ASM/ABM enrolment, when the first Secure Token assignment is vulnerable.

If you are happy with this I'll close and resubmit after updating it. I assume the 89 for cleanupAndExit is for a specific purpose, if you can direct me to the list of error codes, I'll identify a more appropriate one (or someone can provide one).

PicoMitchell:
Ah yes, that prompt to administer the system that `sysadminctl -addUser` will output can by bypassed by the parent process having Full Disk Access. I believe that prompt actually only has to do with UID assignment and if that prompt is declined or not displayed because of a scripted installation without FDA, its failure actually does not prevent the user from being created. I believe the only thing it will do is NOT properly assign a custom UID if one was specified and instead just use the next available UID starting a 501.

I believe it many MDM scenarios Installomator would be running as a child of a process that has FDA, such as the `jamf` binary. And then I believe installer should inherit that FDA access since it would also be a child of a process that has FDA.

Would be worth testing that in a real-world scenario to be certain, or at least invoking Installomator from within Terminal that has been granted FDA to see if the permissions properly propagate down to the installation.

As for exit codes, I don't know the Installomator standards. I'm not an Installomator maintainer, just started poking around with it lately and trying to contribute where I thought I might be able to be helpful.

Also, again, no need close this PR if you want to do any updates, just commit new changes to your branch and they will be applied to this PR automatically.

PicoMitchell:
Also, if you do move forward with creating a user, I think the safest check would be to simply see whether or not a Secure Token user is already present, which could be done with:
```
if [[ "$(diskutil apfs listCryptoUsers /)" == 'No cryptographic users for disk'* ]]; then
    echo 'NO SECURE TOKEN USERS CREATED YET!'
fi
```

If checking for if `_mbsetupuser` was running was important, you could also just use the `currentUser` variable that is already set by Installomator since it is set via the `scutil <<< "show State:/Users/ConsoleUser"` technique which also properly lists `_mbsetupuser` when run at Setup Assistant.

wakco:
Sadly no, Full Disk Access isn't the config in question (There is a specific one for installer, as I discovered the issue with Jamf Pro, which already had FDA at the time :( ). In my testing when I discovered this, I didn't check for a specific UID the package might set, however at the time I stopped checking when I had an account added that Panopto happily worked with. I'll take a closer look to see if there is anything else the package may be setting for the account.

Oh thanks for the PR tip, I wondered how to do that, and I couldn't find it. I'll remember that.

PicoMitchell:
I know from testing that FDA definitely does bypass that "would like to administer your computer" prompt. You can see that yourself when you run `sysadminctl -addUser` in Terminal with and without Terminal being granted Full Disk Access.

The problem could be with the inner workings of the `installer` command if maybe it runs things in a way that breaks out of the process chain so it may not inherit FDA when invoked by a process that has FDA.

PicoMitchell:
Would be interested in seeing the `pre`/`postinstall` that the package is doing to create the user. If it isn't setting a custom UID, it may actually be no issue if the command "fails" with the prompt not getting approved. The user will still get created.

wakco:
Going off-topic, I've checked this with Jamf Pro. Including Jamf Pro, we get 3 ways to check the current username during PreStage (DEP/ASM/ABM) enrolment, `scutil <<< "show State:/Users/ConsoleUser"` returns nothing during enrolment, `who | grep console | cut -d ' ' -f 1` returns `_mbsetupuser`, and Jamf Pro's own `$3` also returns nothing (I suspect it uses the scuttle method).

Back on topic, I'll let you know what I find regarding Panopto's installer package creation of panopto_upload.

PicoMitchell:
Hmmm, interesting. In all my testing the `scutil` method has properly listed `_mbsetupuser` when run at Setup Assistant, but that was always via LaunchDaemon, so maybe something is different with when/how MDM enrollment stuff is being run.

Here's an example of the raw output from the `scutil` technique during Setup Assistant:
```
<dictionary> {
  GID : 248
  Name : _mbsetupuser
  SessionInfo : <array> {
    0 : <dictionary> {
      kCGSSessionAuditIDKey : 100003
      kCGSSessionGroupIDKey : 248
      kCGSSessionIDKey : 257
      kCGSSessionLoginwindowSafeLogin : FALSE
      kCGSSessionOnConsoleKey : TRUE
      kCGSSessionSystemSafeBoot : FALSE
      kCGSSessionUserIDKey : 248
      kCGSSessionUserNameKey : _mbsetupuser
      kCGSessionLoginDoneKey : TRUE
      kCGSessionLongUserNameKey : Setup User
      kSCSecuritySessionID : 100003
    }
  }
  UID : 248
}
```

wakco:
From the log of an enrollmentComplete Jamf Pro policy with a script set to execute "Before":
```
Tue Feb 28 17:31:53 PST 2023
ESCROW  UUID                                                                     TYPE USER
_mbsetupuser console  Feb 28 17:29 
<dictionary> {
  SessionInfo : <array> {
    0 : <dictionary> {
      kCGSSessionAuditIDKey : 100067
      kCGSSessionGroupIDKey : 88
      kCGSSessionIDKey : 258
      kCGSSessionOnConsoleKey : TRUE
      kCGSSessionUserIDKey : 88
      kCGSSessionUserNameKey : unknown
      kCGSessionLoginDoneKey : FALSE
      kCGSessionLongUserNameKey : Unknown User
      kSCSecuritySessionID : 100067
    }
  }
}
Tue Feb 28 17:31:53 PST 2023: 
```
 Script contained:
```
#!/bin/zsh
# for testing only
date
fdesetup list -extended
who
scutil <<< "show State:/Users/ConsoleUser"
echo "`date`: $3"
```

> The problem could be with the inner workings of the `installer` command if maybe it runs things in a way that breaks out of the process chain so it may not inherit FDA when invoked by a process that has FDA.

I think you might be right, as the package uses the dscl command to create the user account, then repeatedly uses dscl to append additional setting for the user account. This comes from a `mk_hidden_user.sh` embedded in the package that `preinstall` calls with `./mk_hidden_user.sh ${PANOPTO_USER} ${PANOPTO_HOME}`.

`mk_hidden_user.sh` checks to see if panopto_upload exists with the correct `NFSHomeDirectory` setting, is so it skips account creation, if the `NFSHomeDirectory` setting is missing it deletes the existing account, before recreating it.
If (re)creating the account it searches through the user ID's from 400 through 499 for an ID that is not in use. It then creates the account using 6 dscl lines to:
- Create the username with the found spare user id as `UniqueID`,
- Set the `PrimaryGroupID`,
- Set the `NFSHomeDirectory` to `/var/panopto` which the following is noted above:
```
        # Next command triggers the warning '"Installer" would like to administer your computer.' on macOS 10.14.2.
        # If the user selects 'no', this script is terminated and immature entry remains.
```
- Set `home` to `/var/panopto`
- Set `UserShell` to `/sbin/nologin`
- Set `this_password_is_disabled`
This is the end of the username creation that may have been skipped, It then checks for, and creates, `/var/panopto` if missing, followed by another dscl command to set `AuthenticationAuthority` to `";DisabledUser;"`.
Then lastly it edits the loginwindow plist to enable `Hide500Users`, then add panopto_upload to `HiddenUsersList`.

So I'll adjust it to check for _mbsetupuser with the who command, and per your suggestion, check a Secure Token exists, but I might use `fdesetup list` over diskutil, as I only need to check if the list in empty, as it requires an actual user to have a Secure Token. This is because the BootStrap Token and iCloud tokens appear in both the diskutil command and in `fdesetup list -extended`, while `fdesetup list` produces a list of username,UUID with no headers that is limited to Tokens with usernames attached, which makes a simpler test, it's also faster.

PicoMitchell:
Very interesting about the different `_mbsetupuser` output, thanks for sharing! That may end up being useful for a bit of code in `mkuser`!

If creating the user does end up being needed, I would recommend just using the exact same `dscl` commands that their script is doing rather than `sysadminctl -addUser` since the latter does a lot of other little things that you don't really want for a "Service Account". But also, I'm curious what part of their code using `dscl` is requiring FDA and triggering that "administer the computer" prompt, because everything I do related to user creation in `mkuser` with uses mainly `dsimport` (and `dscl` for some stuff) and doesn't require FDA and doesn't trigger that "administer the computer" prompt. But, deleting users with `sysadminctl -deleteUser` _or_ `dscl . -delete` does always trigger that prompt. Would it be possible to share the package for me to check out? You could DM it over Slack, if that's possible.

PicoMitchell:
Ah, I see now that you pointed out their comment in the script that it's setting the home folder that triggers the prompt. I'll have to look into that for my own curiosity. `mkuser` does that part with `dsimport` which seems to (luckily) be exempt, maybe because the entire record is being created all at once rather than modifying an existing record like multiple `dscl` commands does.

PicoMitchell:
Doing a little testing with `dscl` commands, and I can only get that "administer your computer" prompt to come up when _modifying_ an existing `NFSHomeDirectory` entry, not when creating it initially for a new user.

So something that their script is doing is likely funky, or maybe that comment is wrong? Have you seen the installer present that prompt yourself?

If that's actively maintained, I'm certain I could make some suggestions that maybe you could pass on to them (since I'm not a customer) to avoid that prompt when creating a new user.

PicoMitchell:
Wait, do they set **both** `NFSHomeDirectory` _and_ `home` attributes? If so, that is their mistake. The latter is simply a synonym for the former, they are both doing the exact same thing. So if that's really what they are doing they are unnecessarily triggering that prompt simply by re-setting the exact same `NFSHomeDirectory` value right after creating it initially.

wakco:
Yes they do...
```
        # Next command triggers the warning '"Installer" would like to administer your computer.' on macOS 10.14.2.
        # If the user selects 'no', this script is terminated and immature entry remains.
        sudo dscl . -append /Users/$username NFSHomeDirectory $homefolder
        sudo dscl . -append /Users/$username home $homefolder
```
That is a direct copy paste, so that shows the order they are set in. I was aware that most adduser tools would do both automatically, and thought they might be linked. I have shared the package, and the script from the package with you over Slack.

I took your advice, and switched to an almost direct copy of what mk_hidden_user.sh was doing, dropping the entry for setting home. I also had added and since removed a copy of it's check for missing NFSHomeDirectory setting where it removes the account, per your comment above about removing triggering a permission request.

PicoMitchell:
Oh jeez, their also using `-append` for the initial creation instead of `-create`. That triggers the "administer your computer" for the first command. If they just did a single `dscl . -create ... NFSHomeDirectory` there would be no prompt and no FDA needed.

Do you think you'd be able to contact their development team? We might be able to get this package script fixed to not even need FDA or hit this prompt. I'm sure they would be happy about it if they knew how easy it was. Most folks just don't spend as much time with `dscl` and friends as I have ended up doing!

wakco:
Oh joy, one more update to the label, and then tomorrow I'll try to provide feedback to panopto.com through work as a customer (they might "hear" it better that way).

And thanks again, I know the feeling regarding hours spent investigating something.

Just to clarify, I should replace all the `-append`'s with `-create`'s, what about that `-passwd`, is that ok as it is?

wakco:
Done, I'll add a comment to the code pending their response.

PicoMitchell:
Yeah, changing all `dscl` `-append` to `-create` commands was correct. Not sure about setting the password since I don't know whether or not the app actually needs that password set. Wouldn't want to break functionality of it does for some strange/bad reason.

Since this user is being made with dscl commands, it could now be allowed to be created as the very first user AS LONG AS you run `dscl . -append /Users/<user name> AuthenticationAuthority ';DisabledTags;SecureToken'` BEFORE setting the password (the `-passwd` command) as is shown in https://support.apple.com/guide/deployment/use-secure-and-bootstrap-tokens-dep24dbdcf9e/web. Also, that command can/should be `-append` in general so that existing `AuthenticationAuthority` values are not overwritten, even though none should exist at that point it doesn't hurt to use `-append` there.

With the account having a UID under 500 it would be guaranteed to never get the first Secure Token on Catalina and older and with that tag added to the AuthenticationAuthority before the password is set it would never get the first Secure Token on Big Sur and newer. So, the conditions to block creation at Setup Assistant or before the first ST is granted could be removed and all would be safe.

With those changes, this code would officially be better/safer than ever allowing the package to create the user.

wakco:
Done, thanks again.

scriptingosx:
I appreciate the effort, but what this label does (create a user) is outside of what I comfortable with the public version of Installomator. I recommend you keep this is a custom label.

wakco:
Ok, I'll remove the add user part, and hope panopto.com will fix their script, after I submit the recommended changes to them.

PR #926, opened 2023-02-28
Label update lansweeper agent options for server and agentkey can be used at the same time

wakco:
Updated, as at least one of lsagentServer and lsagentKey is required, however both can be used.

Thanks to @PicoMitchell for suggestion on code clean up. And I have also cleaned up the formatting and error message (which was also incorrectly referencing lsagentCloudKey instead of lsagentKey).


scriptingosx:
Thank you!

PR #933, opened 2023-03-02
Update Adobe Reader Label to be a single label, use new naming format…

isaacatmann:
Fixes issues with Adobe Reader, merges install/update into a single label.


scriptingosx:
Thank you!

PR #942, opened 2023-03-08
Loom label: Corrected URLs to not be CDN specific

acodega:
Loom CDN changes broke the existing URLs which pointed to specific S3 repositories. I also removed the credit line while I was at it to clean it up.


scriptingosx:
Thank you!

PR #948, opened 2023-03-13
Add labels for python, anaconda, and crashplan

greatkemo:
This is the label to download and install python from python.org


scriptingosx:
Thank you!

In the future, please create a single PR per label.

PR #953, opened 2023-03-15
Update jabradirect.sh

Theile:
New `appNewVersion` that looks at their json file, and not try to isolate it on the release note page.


scriptingosx:
Thank you!

PR #955, opened 2023-03-17
Update wireshark.sh

theadamcraig:
Updated download URL.
appNewVersion is no longer necessary. but i left it commented out.


timnottom:
For what it's worth: I tried this change and it worked.

scriptingosx:
Thank you!

PR #956, opened 2023-03-17
Making the displaynotification function prefer swiftdialog if it's installed

theadamcraig:
Making the displaynotification function prefer swiftdialog if it's installed.


scriptingosx:
see also #1046

PR #960, opened 2023-03-21
Dymoconnectdesktop label

gabemarcelino:
Added Dymo Connect label, downloadURL uses the Dymo Compatibility chart webpage (https://www.dymo.com/compatibility-chart.html ) and looks for all the pkg links and arranges the latest. App New Version looks for the same links and cuts the version number.


scriptingosx:
Thank you!

PR #964, opened 2023-03-23
Update GitKraken label

Maarten-Wijnants:
On intel processors this label always downloaded an older version of GitKraken.


scriptingosx:
Thank you!

PR #967, opened 2023-03-27
Add arcbrowser

Aleexiiiiiii:
Add arc browser to installomator


scriptingosx:
Did you actually test this label with `./assemble.sh arcbrowser`?

If you had, you would have noticed that the filename of your label was missing the necessary `.sh.

Fixed this in merge, thank you!

PR #992, opened 2023-04-14
Add label for Podman Desktop

drewdiver:


drewdiver:
This installs the universal binary of Podman Desktop

PR #996, opened 2023-04-17
Additional download check

Theile:


Theile:
This extra download check will try to detect if a proxy server returned some ASCII code (that will be a html page), and then script will fail if that is the case.

PR #1010, opened 2023-04-21
GPG Suite: Update domain

dnikles:
switch from gpgtools.org to gpgtools.com
The install file is no longer hosted on gpgtools.org


scriptingosx:
duplicate #946

PR #1015, opened 2023-04-27
New label: Patchomator

JordyThery:
Created label for Patchomator (see https://github.com/Mac-Nerd/patchomator).


scriptingosx:
Thank you!

PR #1027, opened 2023-05-03
Combine Itsycal/mowgliiitsycal into one

duderin0:
The contents of these two are exactly the same, but with two different labels. Combining them now.

Also changed the links becuase Itsycal is now a universal app and the referenced xml's between the two labels are now identical, so removing the extra mention of apple silicon to simplify.


scriptingosx:
Thank you!

PR #1031, opened 2023-05-04
Minor cosmetic alignment for `printlog`

Theile:
This minor fix will correct the alignment of of `log_priority` even if if used in repetition.


scriptingosx:
thank you

PR #1036, opened 2023-05-10
Update label: adobereaderdc-install.sh

amacomber:
current curl command reveals incorrect version. updated site and curl to download latest release

### DIFF
--- a/fragments/labels/adobereaderdc-install.sh
+++ b/fragments/labels/adobereaderdc-install.sh
@@ -5,7 +5,7 @@ adobereaderdc-install)
     printlog "Changing IFS for Adobe Reader" INFO
     SAVEIFS=$IFS
     IFS=$'\n'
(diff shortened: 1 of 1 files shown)


isaacatmann:
The reason we get the full versions list is adobe doesn't provide an installer for every version of Acrobat, so the most current version of the acrobat app is possibly not the most current version of the adobe installer. There are cases like this where they add a X into the version because windows has a different version number than man which means we don't always get the latest.. But I like what you're doing here, the following will add both the "most current" and the previous 30 versions to the versions array giving us more options:
```
versions=( $( curl -sL https://armmf.adobe.com/arm-manifests/mac/AcrobatDC/reader/current_version.txt | rev | cut -c1- | rev | sed -e 's/\.//g') )
versions=+( $( curl -s https://www.adobe.com/devnet-docs/acrobatetk/tools/ReleaseNotesDC/index.html | grep -Eo "[0-9]+\.[0-9]+\.[0-9]+"| head -n 30) )
```

Will you also test this?

scriptingosx:
this is a duplicate of at least in conflict with #933 
can you check if #933 solves this, as well?

PR #1053, opened 2023-05-24
Updated label: TeamViewer

kryptonit1:
Fixed appNewVersion variable due to change of URL by TeamViewer.


scriptingosx:
the previous url with the `mac-os` URL still works for me for downloading, so why the change?

scriptingosx:
duplicate of #1054

kryptonit1:
> the previous url with the `mac-os` URL still works for me for downloading, so why the change?

There's no problem with downloading the application. It's just that the `appNewVersion` response is none when using `mac-os` instead of `macos` so comparison with installed version is not working.

> duplicate of #1054

This is not a duplicate of #1054. Different labels with the same `appNewVersion` problem.

PR #1054, opened 2023-05-24
Updated Label: TeamViewer Host

kryptonit1:
Fixed appNewVersion variable due to change of URL by TeamViewer.


scriptingosx:
this one supersedes the change in #1053, right?

kryptonit1:
> this one supersedes the change in #1053, right?

No, these are two different labels (teamviewer and teamviewerhost) with the same problem.

scriptingosx:
Thank you!

PR #1069, opened 2023-05-29
Updated Label: Cytoscape

kryptonit1:
Fixes a problem that causes the Apple Silicon version of Cytoscape to be downloaded to Intel devices with installation failure as result.


scriptingosx:
Thank you!

PR #1071, opened 2023-05-30
Create camtasia2023.sh

beatlemike:
Added camtasia2023 label, this will supersede camtasia2022

### DIFF
--- a/fragments/labels/camtasia2023.sh
+++ b/fragments/labels/camtasia2023.sh
@@ -1,0 +1,7 @@
+camtasia2023)
+    name="Camtasia 2023"
+    type="dmg"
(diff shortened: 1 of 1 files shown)


duderin0:
Do we want to follow the previous convention where the latest version gets the default name without the year as well? Currently for 2022 its:
camtasia|\
camtasia2022)

beatlemike:
I think that makes sense. I noticed the fragment for 2022 is just named "camtasia.sh". So would technically that need to be updated with the camtasia2023 label, and a "camtasia2022.sh" be created?

scriptingosx:
yes, I agree on the file name change. Follow-up question: can we retire some of the labels for the older versions or are those versions of the app still supported?

scriptingosx:
I changed the file and label names in the commit above. Thank you!

(Still expecting feedback on whether we can or should delete old labels here.)

beatlemike:
> yes, I agree on the file name change. Follow-up question: can we retire some of the labels for the older versions or are those versions of the app still supported?

I can't speak to that directly. When a. new version comes out, they do discontinue updating the previous version. But the do make versions back to 2012 available for download. I would say Leaner and Meaner. Maybe drop all but the last 1 or 2 versions?

beatlemike:
and thank you my friend.

PR #1076, opened 2023-06-02
brave label appNewVersion

Mac-Nerd:
in appNewVersion, xpath call requires -e switch


scriptingosx:
it does not: https://github.com/Installomator/Installomator/blob/6b047bcd771780973040166e30387185eb9ea643/fragments/functions.sh#L209

Mac-Nerd:
oh hey, would you look at that. I'll have to dig further to see why it wasn't getting the version.

PR #1077, opened 2023-06-02
added bravepkg.sh

Mac-Nerd:
New PKG installer for Brave Browser, with a single download URL for universal x86/ARM64 installer.


scriptingosx:
Thank you!

PR #1088, opened 2023-06-07
Dropbox with Apple Silicon support

Theile:
Automatically select between architectures. Also `appNewVersion` is fixed for control characters


scriptingosx:
Thank you!

PR #1104, opened 2023-06-14
Update microsoftedge.sh

Theile:
`microsoftedge` label does not work with `autoupdate` anymore, as Microsoft now has introduced a separate updater for Edge.

I post this for others to investigate, as it is not fully working yet. The update command gives errors on my system.

`updateToolArguments` might not be correct.


scriptingosx:
branched from `dev`?

Theile:
Will delete this, as changes are too many for `main`, but it looks like It is working.

PR #1109, opened 2023-06-19
New label TeamViewer Host custom

JordyThery:
Based on the custom TeamViewer QS label adds support for downloading a custom TeamViewer Host version.

### DIFF
--- a/fragments/labels/teamviewerhostcustom.sh
+++ b/fragments/labels/teamviewerhostcustom.sh
@@ -1,0 +1,11 @@
+teamviewerhostcustom)
+    name="TeamViewerHost"
+    type="pkgInDmg"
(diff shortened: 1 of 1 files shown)


scriptingosx:
Not sure how to test this, since I don't have a custom download URL.

This is probably a stupid question, but I have little experience with TeamViewer: how is this different from `teamviewerqscustom`?

JordyThery:
Hi @scriptingosx,

TeamViewer has different versions. QS is the QuickSupport version (only allows the Mac to be remote controlled if the app is open / doesn't work at the login window). The host version allows it running in the background continuously and also allows the login window to be taken over. 

Lastly there's the full version which allows to take control of other devices, that as far as I know does not have a 'custom branded' install.

scriptingosx:
thank you!

PR #1111, opened 2023-06-22
Label 'remotedesktopmanagerfree' URL change

jisgett:
The URL for downloading Remote Desktop Manager Free has changed to include an additional '/' at the end


acodega:
As discussed on Slack, you want to edit the label inside of the fragments folder, not the Installomator script itself. Also put your summary of changes in your commit message (instead of "Update Installomator.sh") not just in the pull request message. :-)

PR #1119, opened 2023-07-08
Label update: libreoffice

macmangr:
Updated Label to install latest stable version (not latest version) and more precise version check


scriptingosx:
Thank you!

PR #1127, opened 2023-07-10
Label fix: VMWareHorizonClient- install type changed

duderin0:
Fix for #1117 

Changed type Dmg to pkgInDmg. Tested with a fresh, full install and it worked; though logs below are only for the assemble.

Output:
./assemble.sh vmwarehorizonclient
2023-07-10 14:21:43 : REQ   : vmwarehorizonclient : ################## Start Installomator v. 10.5beta, date 2023-07-10
2023-07-10 14:21:43 : INFO  : vmwarehorizonclient : ################## Version: 10.5beta
2023-07-10 14:21:43 : INFO  : vmwarehorizonclient : ################## Date: 2023-07-10
2023-07-10 14:21:43 : INFO  : vmwarehorizonclient : ################## vmwarehorizonclient
2023-07-10 14:21:43 : DEBUG : vmwarehorizonclient : DEBUG mode 1 enabled.
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : name=VMware Horizon Client
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : appName=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : type=pkgInDmg
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : archiveName=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : downloadURL=https://download3.vmware.com/software/CART24FQ2_MAC_2306/VMware-Horizon-Client-2306-8.10.0-21964677.dmg
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : curlOptions=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : appNewVersion=8.10.0
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : appCustomVersion function: Not defined
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : versionKey=CFBundleShortVersionString
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : packageID=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : pkgName=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : choiceChangesXML=
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : expectedTeamID=EG7KH642X6
2023-07-10 14:21:46 : DEBUG : vmwarehorizonclient : blockingProcesses=
[...]

PR #1132, opened 2023-07-12
update type to pkgInDMG

opensourceADHD:
Noticed today that the label for this was successfully pulling down the dmg, but was looking for "**VMware Horizon Client.app**" instead of the PKG installer.
Updated the type and confirmed locally.


opensourceADHD:
Oop, closing this, saw someone put a much more detailed PR in the other day for the same thing. 😅 #1127

cnita2:
I've tried to test this new label and doesn't seem to work. Anyone else got it working?

PR #1140, opened 2023-07-18
New Label: GraphPad Prism 10

kryptonit1:
2023-07-18 08:30:13 : REQ   : prism10 : ################## Start Installomator v. 10.4, date 2023-06-02
2023-07-18 08:30:13 : INFO  : prism10 : ################## Version: 10.4
2023-07-18 08:30:13 : INFO  : prism10 : ################## Date: 2023-06-02
2023-07-18 08:30:13 : INFO  : prism10 : ################## prism10
2023-07-18 08:30:13 : INFO  : prism10 : SwiftDialog is not installed, clear cmd file var
2023-07-18 08:30:15 : INFO  : prism10 : setting variable from argument DEBUG=0
2023-07-18 08:30:15 : INFO  : prism10 : BLOCKING_PROCESS_ACTION=tell_user
2023-07-18 08:30:15 : INFO  : prism10 : NOTIFY=success
2023-07-18 08:30:15 : INFO  : prism10 : LOGGING=INFO
2023-07-18 08:30:15 : INFO  : prism10 : LOGO=/System/Applications/App Store.app/Contents/Resources/AppIcon.icns
2023-07-18 08:30:15 : INFO  : prism10 : Label type: dmg
2023-07-18 08:30:15 : INFO  : prism10 : archiveName: Prism 10.dmg
2023-07-18 08:30:15 : INFO  : prism10 : no blocking processes defined, using Prism 10 as default
2023-07-18 08:30:15 : INFO  : prism10 : name: Prism 10, appName: Prism 10.app
2023-07-18 08:30:15.771 mdfind[7302:100296] [UserQueryParser] Loading keywords and predicates for locale "en_US"
2023-07-18 08:30:15.771 mdfind[7302:100296] [UserQueryParser] Loading keywords and predicates for locale "en"
2023-07-18 08:30:15.811 mdfind[7302:100296] Couldn't determine the mapping between prefab keywords and predicates.
2023-07-18 08:30:15 : WARN  : prism10 : No previous app found
2023-07-18 08:30:15 : WARN  : prism10 : could not find Prism 10.app
2023-07-18 08:30:15 : INFO  : prism10 : appversion:
2023-07-18 08:30:15 : INFO  : prism10 : Latest version of Prism 10 is 10.0.0
2023-07-18 08:30:15 : REQ   : prism10 : Downloading https://cdn.graphpad.com/downloads/prism/10/InstallPrism10.dmg to Prism 10.dmg
[...]

PR #1141, opened 2023-07-18
Updated Label: GraphPad Prism 9

kryptonit1:
GraphPad Prism 10 has been released (https://www.graphpad.com/updates/prism-1000-release-notes) and this has caused the current appNewVersion variable for the GraphPad Prism 9 label to show version 10.0.0 as the latest version, which is a new major version. It's unlikely that Prism 9 will receive any more updates so I've changed the appNewVersion label to statically be the last released version of Prism 9 (9.5.1) and also created a PR for a Prism 10 label (#1140).

### DIFF
--- a/fragments/labels/prism9.sh
+++ b/fragments/labels/prism9.sh
@@ -2,6 +2,6 @@ prism9)
     name="Prism 9"
     type="dmg"
     downloadURL="https://cdn.graphpad.com/downloads/prism/9/InstallPrism9.dmg"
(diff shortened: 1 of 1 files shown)


scriptingosx:
I don't like that you are hardcoding the appNewVersion here. Are you absolutely certain this is the ultimate version? If so, why keep the label?

scriptingosx:
Thank you!

PR #1145, opened 2023-07-19
Update dropbox.sh label

delize:
Adds support for Apple Silicon installer that can be downloaded with a ?arch={arm64|intel}, depending on the device.

I specified x86_64 versus i386, if this is incorrect do let me know.


scriptingosx:
duplicate #1088

PR #1148, opened 2023-07-24
Update Label: Slack

guy-tallon:
Added the check to download the Intel or Apple Silicon versions of Slack. Half the download size


scriptingosx:
Thank you!

PR #1150, opened 2023-07-25
New Label: Sassafras KeyAccess client for AllSight (formerly KeyServer)

wakco:
Adding a label for the KeyAccess client, as they now have a cloud hosted option (and we are in the process of moving ours from hosted to cloud).

(Sorry for the repost, removed the , from the branch name without realising it would close the pull request.)


scriptingosx:
Thank you!

PR #1155, opened 2023-07-28
Update nextcloud Label

northalpha:
It looks like starting version 3.9 nextcloud is not publishing the releases packages on their main nextcloud desktop repo anymore but have a separate one for releases only, but this one is not in the nextcloud organisation on github. but i have checked the package, signature and developer ID / notarization and the github releases are signed by a valid and nextcloud org commit able person.


scriptingosx:
Thank you!

PR #1173, opened 2023-08-18
Git tag support

bartreardon:
I had a crack at  `downloadURLFromGit` and adjacent code to see if I could get tag support in there. The idea being you could specify a git tag with the label and you get that version, regardless what the latest is.

I have it working without too much hassle (and discovered the 60 request limit per hour for unauthenticated api requests) but there's a minor issue. I thought I could have it take an argument like `GIT_TAG=v1.2.3` but the code for processing those arguments ends up after all the labels so when called, the label doesn't end up using the tag value and will default to fetching latest instead.

Some ideas:

1) There's nothing for $2 so that could be used for github tags but I'm not sure about using it since github is but one option for download source and hard coding $2 to something like that doesn't seem ideal. This would also be inconvenient for existing implementations as admins would have to move any other arguments they are using in order to accomodate `$2`

2) Insert the version number into the label name with a unique separator character and extract it from `$1` if present. `-` and `_` are not available as there are labels that use those. Some other non command character like `#` might work.

```bash
label_string="$1"
parts=("${(@s[#])label_string}")
label="${parts[1]}"
GIT_TAG="${parts[2]}"
```

To use, one would just pass in a label in the format  `somelabel#<tag>`, e.g. `swiftdialog#v2.2.1` . Not entering a tag works the same as it always has 

I've implemented this latter form and it appears to work well but open to discussion.

One concern with the code a the moment is the number of requests to the github API which with this code uses 5 requests. Un-authenticated access is limited to 60 requests per hour so there would be a cap of 12 installs from github sources before the limit was reached. Given the use in onboarding workflows I can see scenarios where this limit would be reached fairly quickly. With that in mind, I'll be looking at reducing the number of calls to the github api


bartreardon:
This works for getting the latest tag without expending api requests

`curl -sLI "https://github.com/$gitusername/$gitreponame/releases/latest" | grep -i "^location" | tr "/" "\n" | tail -1`

fragments/functions.sh
```diff
@@ -166,17 +175,19 @@ downloadURLFromGit() { # $1 git user name, $2 git repo name
         filetype=$type
     fi
 
+    gitURL="$gitusername/$gitreponame/releases/tags/$gitReleaseTag"
+
     if [ -n "$archiveName" ]; then
-        downloadURL=$(curl -sfL "https://api.github.com/repos/$gitusername/$gitreponame/releases/latest" | awk -F '"' "/browser_download_url/ && /$archiveName\"/ { print \$4; exit }")
+        downloadURL=$(curl -sfL "https://api.github.com/repos/$gitURL" | awk -F '"' "/browser_download_url/ && /$archiveName\"/ { print \$4; exit }")
```
bartreardon, line 181: The fallback method which doesn't use the API works fairly well. Any reason to prefer a call to the API first? (other than bandwidth considerations).

got API requests down to 3 though so one would have to install 20 github sourced packages to it the limit

PR #1183, opened 2023-08-24
New Label: Acorn image editor

bartreardon:
Adds label for the [Acorn image editor](https://flyingmeat.com/acorn/)

This software is non-free but has a 14 day trial. Also as far as I can tell the download is the same as the update


scriptingosx:
Thank you!

PR #1185, opened 2023-08-24
New Labels: Chemdoodle and Chemdoodle 3D

bartreardon:
ChemDoodle is 2d and 3d drawing application for molecules and chemistry https://www.ichemlabs.com

uses the `folderName` property to copy from a folder within a DMG


scriptingosx:
Thank you!

PR #1188, opened 2023-08-25
New Label: texshop

bartreardon:
Label for TeXShop obtained from the unofficial TeXShop source code repository on github (which is linked to from the original sources https://pages.uoregon.edu/koch/texshop/obtaining.html)

It doesn't appear to be in lock step with that source currently being one version behind however it's a much more reliable source to download from


scriptingosx:
Thank you!

PR #1189, opened 2023-08-25
New Label: bibdesk

bartreardon:
edit and manage your bibliography

https://bibdesk.sourceforge.io

also sourceforge is a PITA to source from.


scriptingosx:
Thank you!

PR #1193, opened 2023-08-28
Add Wondershare UniConverter 15

gabemarchan:
Wondershare is infamous for not having permanent links to the latest version of their software, so you need to reach out to their support team every few weeks to get the newest one.


scriptingosx:
The point of Installomator is to _automate_ the download and installation. If you have to manually update the url frequently, I would recommend using `valuesfromarguments` instead

PR #1198, opened 2023-09-01
tunnelblick.sh to use beta

bbrendon:
Please review. This is a little janky but its the bets I could come up with.


scriptingosx:
Thank you!

PR #1204, opened 2023-09-07
Create libericajdk8ltsfull.sh

drewdiver:
We are beginning to use Liberica JDK 8 LTS Full at my company and I thought I would share the label I made to allow use with Installomator.

Their JDK 8 specifically is erm "Special" in that the version in the package is condensed to 8u382+6 while the `/Library/Java/JavaVirtualMachines/liberica-jdk-8-full/Contents/Info.plist` reveals `1.8.0_382-b06`. The other LTS versions match.

They have an open API to query the latest where you receive a JSON payload. Using one call I can get the bits of info I need and match a section of the `CFBundleVersion` to the `updateVersion` from the JSON.

Using this string we can then get a download link without further JSON parsing.

I would have used their GitHub release page but they lump all various versions (8 / 11 /17 / 20) together. I feel by using their API, I can ensure the correct link.

### DIFF
--- a/fragments/labels/libericajdk8ltsfull.sh
+++ b/fragments/labels/libericajdk8ltsfull.sh
@@ -1,0 +1,22 @@
+libericajdk8ltsfull)
+    name="Liberica JDK 8 Full LTS"
+    type="pkg"
+    case $(arch) in
+        "arm64")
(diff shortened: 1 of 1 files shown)


scriptingosx:
Thank you!

PR #1207, opened 2023-09-07
Libericajdk20full

drewdiver:
A label for getting the latest Liberica JDK 20 pkg


scriptingosx:
ok, why are you using dashes in the label name for the lts version? this is inconsistent with all the previous labels, and the non-lts 20 label?

PR #1211, opened 2023-09-13
Label vectorworks2021sp5

TacoDienstag:
Full log on: 
https://gist.github.com/TacoDienstag/53422fc4834902ebe7cfdab11282ebed


scriptingosx:
this is a service patch with a static download url. I would rather not create a label for this. You can use Installomator to download and install this using `valuesfromarguments`

PR #1212, opened 2023-09-14
Create JamfCheck label (new app)

JordyThery:
Creates a label for Jamf Check (see https://github.com/txhaflaire/JamfCheck).


scriptingosx:
Thank you!

PR #1213, opened 2023-09-14
Update canva.sh

Odnlinx:
The arm version of the installer is no longer being updated. There is an Universal installer now.


scriptingosx:
I get being cautious, but we can always restore to previous version using the git history. Please delete the commented lines

scriptingosx:
Thank you!

PR #1220, opened 2023-09-25
Update Progress 1st swiftDialog.sh to use native checkmark

mwmatter:
Updating the installed checkmark to use the native swift dialog "success" status instead of the emoji checkmark. 

There are probably other Swift Dialog files that need to be updated to match this change.


Theile:
I'm including this change in my updated MDM scripts in #1219. Thank you for this!

PR #1226, opened 2023-09-29
Update zohoworkdrivetruesync.sh

Theile:
If only I kept the original found `downloadURL`, it would still work. Hope it continues to be this strange URL.


scriptingosx:
Thank you!

PR #1240, opened 2023-10-06
add imagr support

keeleysam:


scriptingosx:
imagr hasn't been updated in 4 yours and probably won't be. Is this really necessary?

PR #1241, opened 2023-10-06
Update python.sh

Mac-Nerd:
previous archiveName returned multiple lines in the result, which broke downloadURL. New grep line returns only the first result.

addresses issue #1232


scriptingosx:
Thank you!

PR #1249, opened 2023-10-10
Create labels for Processing

wakco:
Installs the Processing app from Github and resolves a version matching issue. Covers both version 3.* and 4.* as they are in separate github repos.


scriptingosx:
Thank you!

PR #1251, opened 2023-10-12
Microsoft Teams classic and new

Theile:
`microsoftteams` label is now also `microsoftteamsclassic` as Microsoft has chosen to rename the app and the Helper.

Added `microsoftteamsnew` to point to Teams version 2. (This label might not be correct for `msupdate`app, as I would guess that it should be `TEAM02` instead, but cannot locate what it should be.

Some time in the future, we will move `microsoftteams` from classic to new…


scriptingosx:
Thank you!

PR #1258, opened 2023-10-13
Clarifying log entries about app reopen and blocking process behavior…

acodega:
… when updateTool is used

Code comments and log entries make it unclear that BLOCKING_PROCESS_ACTION is ignored when a label uses updateTool. Also clarified log entry regarding reopening apps.


scriptingosx:
thank you!

PR #1266, opened 2023-10-19
Update label 4kvideodownloader and create labels for 4k's other apps

wakco:
Lebel for 4kvideodownloader was out of date and had stopped working, so here's an updated label that works, and 7 new labels for 4K's other apps.


BigMacAdmin:
It appears that in the time since this pull request the URLs may have changed again, I am getting empty `downloadURL` variables when testing these labels.

PR #1273, opened 2023-10-23
Update bettertouchtool.sh

robjschroeder:
Updated the `appNewVersion` variable as it was looking at a page that was not being kept up to date by the vendor.


BigMacAdmin:
Rob,

When testing this I noticed a discrepancy between the `appNewVersion` value and the actual version of the app being installed.

`appNewVersion` resulted in 4.426 but the actual downloaded app was version 4.403

PR #1274, opened 2023-10-23
Fixes to 10.5

Theile:
Various fixes and changes to labels, like `appNewVersion`.

### DIFF
--- a/fragments/disabled/fontexplorer.sh
+++ b/fragments/disabled/fontexplorer.sh
@@ -1,4 +1,5 @@
 fontexplorer)
+    # Product converted to https://www.monotypefonts.com/
     name="FontExplorer X Pro"
     type="dmg"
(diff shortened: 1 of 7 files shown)


Theile:
Fixes #1289.

scriptsandthings:
@Theile the duplicate label for microsoftonedrive-rollingout still appears in the labels.txt on the master (dev) branch https://github.com/Installomator/Installomator/blob/main/Labels.txt

I notice when running App Auto-Patch, although i have microsoftonedrive-rollingout listed as an label to ignore, it's still found on every run of the policy (latest version of App Auto-Patch and Installomator). 

The labels.txt probably isn't the cause of my issue, but since I notice it was listed twice there I thought I'd mention it.

Theile:
Yes, and you need the corrections from this PR.

PR #1288, opened 2023-10-25
update notion label

alex02th:
Now there is only one url to download the universal binaries for mac

assemble.sh passed


BigMacAdmin:
It looks like it took us long enough to process this that Notion now has a universal app.

Do you want to try rewriting this using this URL? `https://www.notion.so/desktop/mac-universal/download`

BigMacAdmin:
Thank you!

PR #1293, opened 2023-10-27
Update splashtopbusiness.sh

jcmbowman:
Fix downloadURL to follow redirect and provide actual downloaded filename. This also fixes issue with not properly determining appNewVersion.


BigMacAdmin:
Thank you!

PR #1294, opened 2023-10-27
new label:logitune

cchambers-cap:
Created a label for Logitech LogiTune software and added logitune to the Labels.txt file. LogiTune simplifies headset and webcam control.

I was not able to test the label because it wasn't found in the labels.txt file when using the assemble.sh command (I'm new to this but assume the label has to exist first?)

### DIFF
--- a/Labels.txt
+++ b/Labels.txt
@@ -343,6 +343,7 @@ linear
 logioptions
 logitechoptions
 logitechoptionsplus
+logitune
 logseq
 loom
 lowprofile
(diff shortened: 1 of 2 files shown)


Theile:
You don't have to put anything into labels.txt, that file is auto-generated when we publish a new version. 

In order to test a label, you can use assembly.sh in utils+folder.

cchambers-cap:
Updated file tested with results in the comments for the commit

BigMacAdmin:
Thank you for your contribution.

I am closing this in favor of #1424 which is more feature complete.

PR #1297, opened 2023-10-31
new label: yubikeymanager

reuven-cohen:
Label for YubiKey Manager CLI Installer.
This installer is not included in the QT version thus requires its own label.


BigMacAdmin:
Thank you!

PR #1299, opened 2023-10-31
Update and clean up microsoft one drive labels

wakco:
microsoftonedrive-deferred.sh actually contained the same label as microsoftonedrive-rollingout.sh, both the rollingout files had MAU code commented out, and since the only things different between all four labels was the download link and if the MAU code was needed, I thought it might be better to simplyfy it all into a single label that accounts for the differences and fixes the label error in microsoftonedrive-deferred.sh.


Theile:
This was already fixed in #1274

PR #1304, opened 2023-11-01
add label xeroxprintandscan

robotprom:
installs drivers for many of Xerox's printers and AIO machines

### DIFF
--- a/fragments/labels/xeroxprintandscan.sh
+++ b/fragments/labels/xeroxprintandscan.sh
@@ -1,0 +1,6 @@
+xeroxprintandscan)
+    name="xeroxprintandscan"
+    type="pkgInDmg"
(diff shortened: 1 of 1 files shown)


rougegoat:
I believe this is already implemented using the xeroxphaser7800 and xeroxworkcentre7800 labels.

robotprom:
> I believe this is already implemented using the xeroxphaser7800 and xeroxworkcentre7800 labels.

well, change your label name to be generic for all Xerox printers in nature and I'll close this pull request. It seems like no matter which Xerox model you select, it points you to same package to install.

BigMacAdmin:
@robotprom One label can be called by multiple names. See `fragments/labels/snagit.sh` for an example of how to accomplish this.

I'm going to close this pull request, if you would like to open one to make the xerox name more generic please do so. Apologies for the delay in updating this PR.

PR #1328, opened 2023-11-13
Create gosign.sh

lorenzo-allevi:
adding a Label for Gosign, (https://www.infocert.it/firma-infocert-it/prodotti/gosign/),

An APP for sign with digital sign / certificate documents or file. This App is very common in italy.
AppNewVersion is not avaiable because there aren't documentation on developer site.

### DIFF
--- a/fragments/labels/gosign.sh
+++ b/fragments/labels/gosign.sh
@@ -1,0 +1,7 @@
+gosign)
+    name="GoSign-Desktop"
+    type="dmg"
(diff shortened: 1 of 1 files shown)


BigMacAdmin:
Thanks for your patience.

It looks as though the `downloadURL` no does not work. Perhaps something changed between now and when you submitted this. Could you please review?

lorenzo-allevi:
hi,
I tried the downloadurl just now and successfully downloaded the file "GoSign-Desktop-installer-osx.dmg" can you retry this?
Perhaps the server are not available when you have try this label?

PR #1329, opened 2023-11-13
Create namiral.sh

lorenzo-allevi:
adding a Label for Firmacerta, (https://www.firmacerta.it),
An APP for sign with digital sign / certificate documents or file. This App is very common in italy. AppNewVersion is not avaiable because there aren't documentation on developer site.


BigMacAdmin:
Thank you!

PR #1346, opened 2023-11-22
Update virtualbox.sh

timnottom:
The VirtualBox label/fragment now downloads from the Japanese URL and is returning an old version: 7.0.10, while the URL without a country specified is giving the latest version 7.0.12.

We could change the url from:
https://www.oracle.com/jp/virtualization/technologies/vm/downloads/virtualbox-downloads.html
to
https://www.oracle.com/virtualization/technologies/vm/downloads/virtualbox-downloads.html

### DIFF
--- a/fragments/labels/virtualbox.sh
+++ b/fragments/labels/virtualbox.sh
@@ -8,7 +8,7 @@ virtualbox)
     elif [[ $(arch) == arm64 ]]; then
         platform="macOSArm64"
     fi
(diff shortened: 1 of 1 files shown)


BigMacAdmin:
This label is not working for me. It looks like VirtualBox no longer has separate downloads for Intel vs. Silicon

I'm closing this in favor of #1357 which removes the intel vs. arm logic

PR #1357, opened 2023-12-01
virtualbox has no ARM version

Mac-Nerd:
There's no ARM/Intel specific versions of Virtualbox. Previous version failed on ARM Macs.


BigMacAdmin:
Thank you!

PR #1363, opened 2023-12-06
Label: arduinoide

jcmbowman:
Label for installing Arduino IDE


BigMacAdmin:
Thank you for your contribution.

I am closing this in favor of duplicate #1505 only because that label uses the Github repo, which may be more reliable than parsing the URL of the developer.

PR #1366, opened 2023-12-08
Update googlechromepkg.sh

jshtux:
omahaproxy is permanently shutdown; must use google api service to get current version now

PR #1367, opened 2023-12-08
Update googlechromeenterprise.sh

jshtux:
omahaproxy is permanently shutdown; must use google api service to get current version now


BigMacAdmin:
Does it make sense to use similar or identical logic to the discussion on #1366 here on this label? If so, could you please update your Pull Request? @jshtux  @fraserhess

jshtux:
this pull request can be cancelled

scriptingosx:
can we use the same appNewVersion as in #1569 ?

PR #1387, opened 2023-12-21
Added a label for FontBase

eumelkind:
Added a label for FontBase (http://www.fontba.se).


eumelkind:
Developer provided https://releases.fontba.se/mac/latest-mac.yml containing current version information upon request via e-mail.

Theile:
We prefer to not have credit lines. As we talked in the team about this, we generally thought the script was long enough, and started to remove these lines…

Anyway we thank you for your contribution, it's highly appreciated!

PR #1405, opened 2024-01-03
Update webex.sh

jshtux:
this gets the latest release version from the release notes page


BigMacAdmin:
Hi, Thank you, but it appears that `appNewVersion` is not working in this Pull Request. I know it's been a while since you have submitted it, but if you have time then please take another look at it.

jshtux:
I updated the appNewVersion directive, I was able to get the current version

BigMacAdmin:
The existing label is working for me as-is.

Is there a compelling reason to change this?

PR #1410, opened 2024-01-08
Enable "-" in versionFromGit

pmex:
Previously 1.2-3 would be read as 1.23, making it hard for some git projects to be read correctly (eg. Swift dialog)


pmex:
Closed with the same reasoning as in #1410

PR #1411, opened 2024-01-08
Modified dialog.sh

pmex:
This will only work with the separate commit #1410 for functions.sh that parses Git versions containing "-"

### DIFF
--- a/fragments/labels/dialog.sh
+++ b/fragments/labels/dialog.sh
@@ -4,6 +4,6 @@ swiftdialog)
     type="pkg"
     packageID="au.csiro.dialogcli"
     downloadURL="$(downloadURLFromGit swiftDialog swiftDialog)"
-    appNewVersion="$(versionFromGit swiftDialog swiftDialog)"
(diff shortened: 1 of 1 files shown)


Theile:
I have not looked much into this, but ou suggest to include "-" in the version number from GitHub, but at least for this label, we need it removed again, as it will not mach?

The #1410 could destroy several other labels like that, so what software have you met that actually requires #1410 to be implemented? Which labels today are failing without this, which kind of GitHub software are we missing because of this?

I'm much more likely to think, that for GitHub software that needs the "-", it would have to be reflected in the label itself, or a different `versionFromGitDash`-function should be implemented to handle this instead.

pmex:
I created the PR because we had a problem with swiftdialog being redownloaded over and over again.
It seems like this has been "fixed" in the later releases of Swift Dialog.
Pre 2.4, they had the build number in the tag. i.e.:
https://github.com/swiftDialog/swiftDialog/releases/tag/v2.3.3-4734

Since this was the only label we use that had this kind of problem, and it seems to be fixed externally, I can't really argument for this to be merged.

Theile:
Super! (Yes, at some time there was an error in the release of SwiftDialog).

PR #1420, opened 2024-01-15
Arguments declared before using

pijpe00:
The DEBUG variable is used before it is declared (as an argument). For example the DEBUG variable in the check root function. By changing the reading arguments place in the script this problem is solved.


Theile:
Yes, but it introduces the problem, that you cannot parse label variables as arguments to the script in order to fix something when you run it. The reading of variables is placed here for a good reason, as they have to be set AFTER the case statement for all the labels.

PR #1423, opened 2024-01-15
Update rancherdesktop.sh

timnottom:
The latest 2 releases don't have "-mac" in the name, causing the installation to fail.

### DIFF
--- a/fragments/labels/rancherdesktop.sh
+++ b/fragments/labels/rancherdesktop.sh
@@ -2,10 +2,10 @@ rancherdesktop)
     name="Rancher Desktop"
     type="zip"
     if [[ $(arch) == "arm64" ]]; then
-      archiveName="Rancher.Desktop-[0-9.]*-mac.aarch64.zip"
+      archiveName="Rancher.Desktop-[0-9.]*.aarch64.zip"
       downloadURL="$(downloadURLFromGit rancher-sandbox rancher-desktop)"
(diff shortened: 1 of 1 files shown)


alexfinn (commented):
The filetype has changed from .zip to .dmg also, the correct label should be 

rancherdesktop)
    name="Rancher Desktop"
    type="dmg"
      archiveName="Rancher.Desktop-[0-9.]*.aarch64.dmg"
      downloadURL="$(downloadURLFromGit rancher-sandbox rancher-desktop)"
      archiveName="Rancher.Desktop-[0-9.]*-.x86_64.dmg"
      downloadURL="$(downloadURLFromGit rancher-sandbox rancher-desktop)"
    fi
    appNewVersion="$(versionFromGit rancher-sandbox rancher-desktop)"
    expectedTeamID="2Q6FHJR3H3"
    appName="Rancher Desktop.app"
    ;;

timnottom:
@alexfinn is dmg better than zip? (genuine question)

timnottom:
It seems that the Rancher Desktop build for version 1.12 was the only one that not contained the "-mac" later versions did have the "-mac" so this pull request can be closed. No action is needed (except maybe the change to dmg as mentioned by @alexfinn if that's neccessary.

Theile:
So will you fix that, or should we close it?

timnottom:
I don't think theres a need for a fix. It seems to work for me as is. You can close it.

PR #1424, opened 2024-01-15
Create logitune.sh

timnottom:
Hi, this label is based on logitechoptionsplus. The main difference in this is that the --quiet of Logitech Options plus isn't in the LogiTune installer app. It will therefore open an additional window where a user need to click install. This isn't the biggest issue when a user decides to initiates the installation, but it is an issue when automatically installing (for example with an update). If someone has a way to check if there are other flags that we can add to make it silent, then I'm all in for it. For us..this label works, but I can imagine that in other cases this would suffice.


rmckellar:
You can add the following line to make it a silent install:

CLIArguments="-silent"

timnottom:
Thanks @rmckellar, that works! I've updated the pull request.

BigMacAdmin:
Thank you!

PR #1428, opened 2024-01-16
New label: Node.js LTS version

meschwartz:
The Node.js website states that the LTS version is "Recommended for Most Users", but Installomator only has a label that installs the "Current" version. This adds a new label for the LTS version.


meschwartz:
Uh oh, I'm not sure what I did wrong here to cause all those old commits to show up in this PR. And I'm not sure the best way to fix it.

meschwartz:
Created a new PR, and closing this one.

PR #1450, opened 2024-01-25
Update bravepkg.sh

kryptonit1:
The appNewVersion variable is broken in the current label. This change fixes the problem.


scriptingosx:
Thank you!

PR #1451, opened 2024-01-26
Update 1password8.sh to remove 1Password Safari from the list

gilburns:
Removed 1Password Safari from the blocking process list. It technically does not block updating the app. Safari should probably be re-launched after the update, but it doesn't prevent the update to the 1Password app


gilburns:
The 1Password 8 extension for Safari is now delivered via a separate AppStore app, not bundled in the 1Password app

https://apps.apple.com/us/app/1password-for-safari/id1569813296?mt=12

BigMacAdmin:
Thank you!

PR #1452, opened 2024-01-26
Update popsql.sh to support arm64 and i386 builds

gilburns:
Updated downloadURL to support arm64 and i386 builds


scriptingosx:
Thank you!

PR #1239, opened 2023-10-05
Label Update: some microsoft products - appNewVersion

kenchan0130:
fix #1237 

<details>
<summary><code>./assemble.sh microsoftexcel</code></summary>

```sh
2023-10-05 23:27:44 : REQ   : microsoftexcel : ################## Start Installomator v. 10.5beta, date 2023-10-05
2023-10-05 23:27:44 : INFO  : microsoftexcel : ################## Version: 10.5beta
2023-10-05 23:27:44 : INFO  : microsoftexcel : ################## Date: 2023-10-05
2023-10-05 23:27:44 : INFO  : microsoftexcel : ################## microsoftexcel
2023-10-05 23:27:44 : DEBUG : microsoftexcel : DEBUG mode 1 enabled.
2023-10-05 23:27:45 : DEBUG : microsoftexcel : name=Microsoft Excel
2023-10-05 23:27:45 : DEBUG : microsoftexcel : appName=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : type=pkg
2023-10-05 23:27:45 : DEBUG : microsoftexcel : archiveName=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : downloadURL=https://go.microsoft.com/fwlink/?linkid=525135
2023-10-05 23:27:45 : DEBUG : microsoftexcel : curlOptions=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : appNewVersion=16.77.23091703
2023-10-05 23:27:45 : DEBUG : microsoftexcel : appCustomVersion function: Not defined
2023-10-05 23:27:45 : DEBUG : microsoftexcel : versionKey=CFBundleShortVersionString
2023-10-05 23:27:45 : DEBUG : microsoftexcel : packageID=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : pkgName=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : choiceChangesXML=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : expectedTeamID=UBF8T346G9
2023-10-05 23:27:45 : DEBUG : microsoftexcel : blockingProcesses=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : installerTool=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : CLIInstaller=
2023-10-05 23:27:45 : DEBUG : microsoftexcel : CLIArguments=
[...]

PR #1455, opened 2024-01-26
Improved handling of `msupdate` errors and no updates available

Theile:
If `msupdate` cannot talk to MAU, `updateTool` will not be used.

`msupdate` will give an “XPC”-error when a MDM solution have not been allowed to initiate communication between `msupdate` and MAU.

Here is the solution for Jamf Pro: https://github.com/pbowden-msft/MobileConfigs/tree/master/Jamf-MSUpdate

Curretnly we have me and a collegue have tested this in Mosyle and Addigy, and even if we deploy similar profiles for them, `msupdate` still gives an error. A workaround might be to stopp the LaunchAgent and start it again, and then it should run, but that is also not pretty, so then Installomator can handle the update if any are available.


scriptingosx:
overlapping changes with #1239, merge conflicts with main branch, please resolve before we merge this into

you can resolve the merge conflict by merging main into this pr branch

PR #1467, opened 2024-02-06
Update thunderbird_intl.sh

kryptonit1:
This would be a nice addition for update workflows since this would check the actual language/locale of the installed Thunderbird app (if installed) before reverting to checking the macOS locale if the Thunderbird app is not installed.

### DIFF
--- a/fragments/labels/thunderbird_intl.sh
+++ b/fragments/labels/thunderbird_intl.sh
@@ -3,7 +3,7 @@ thunderbird_intl)
     # and install corrosponding version of Thunderbird
     name="Thunderbird"
     type="dmg"
-    userLanguage=$(runAsUser defaults read .GlobalPreferences AppleLocale | tr '_' '-')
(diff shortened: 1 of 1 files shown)


scriptingosx:
@Theile what do you think about this?

Theile:
@scriptingosx for firefox we made separate labels for the localized. Often we want to have the us version, so that option is needed. So a separate label named _intl or something…

kryptonit1:
> @scriptingosx for firefox we made separate labels for the localized. Often we want to have the us version, so that option is needed. So a separate label named _intl or something…

This PR is an update for the already existing **thunderbird_intl** label. So this is not a new label.

The difference between the current label and my update is that with my update the label checks the locale/language of Thunderbird (if already installed on the device) and downloads the latest version of Thunderbird with the same language which is already installed. This is useful if the label is used to update already installed versions of Thunderbird. If Thunderbird is not installed, then the label reverts to downloading Thunderbird with the same language as the macOS language on the device (if available), which the current label does.

Theile:
Cool. Let's go with that!

PR #1468, opened 2024-02-07
EPOS Connect

Theile:
Installation of EPOS Connect.

It updates to a version that cannot be found for download, but at least it can instal the software.


BigMacAdmin:
Thank you!

PR #1472, opened 2024-02-08
Create loupedeck.sh

Theile:


BigMacAdmin:
Theile: The `downloadURL` on this label needs some work. It's giving more html data than just the desired URL.

PR #1476, opened 2024-02-13
Mdm script SwiftDdialog fix

Theile:
This PR fixes the MDM scripts so that the call to SwiftDialog is done to the binary in `/usr/local/bin/dialog`.

The Jamf scripts have been testet, but not the other MDM scripts.


wakco:
Just worth noting, `/usr/local/bin/dialog` is not a binary, it's a script, the binary is `/Library/Application Support/Dialog/Dialog.app/Contents/MacOS/Dialog`, I believe the correct note would be directing the scripts to use the developer recommend path of `/usr/local/bin/dialog`, over using the App bundle directly.

I'm not a reviewer, but I hope they see this to better understand @Theile's request, and not get confused by the wording of the request.

Theile:
Yes, thank you. You are absolutely right. We have full focus on getting this merged and released soon.

Issue #1460, opened 2024-01-31
Update Autodesk Fusion 360 Admin Install Label


LukeTommo:
Could someone please update the Autodesk Fusion 360 Admin Install label with the following URL:

https://dl.appstreaming.autodesk.com/production/installers/Autodesk%20Fusion%20Admin%20Install.pkg

acodega:
Fixed in 10.6

PR #1477, opened 2024-02-13
Update autodeskfusion360admininstall.sh

moojomoore:
Update Fusion 360 Admin Install - The previous URL was redirecting to the Fusion 360 Client Install.

Verified appNewVersion is still correct.

In regards to #1460


BigMacAdmin:
Thank you!

engrziia (commented):
Hi,

We were using Fusion for sometime and recently found it is not working anymore.
I tried to change the line five in our installometer code but no luck.
Any suggestions?

PR #1483, opened 2024-02-14
Delete redeye.sh

drewdiver:
Red Eye has moved behind a paywall, see https://www.hexedbits.com/redeye/ where you are directed to Gumroad for purchase.


scriptingosx:
Thank you!

PR #1490, opened 2024-02-16
Extended version checking, MAU availability and update package selection of select Microsoft labels

croaker-1:
Altered the behavior of select `microsoft` labels to check for full version numbers and compare them against both full installer package versions and update (full or delta) package versions. Update packages are often more current than the full installer packages available at any given time. This will prevent repeated attempts to update apps that are already at a version above what is provided by the installer. This can be the case when the app has previously been updated by Microsoft Auto Updater (MAU) or by other means.

Here is a real world example example of the version differences (between installer and updater packages) for Outlook from earlier this month:
Latest version parsed from the installer (default Installomator behavior):
```
https://go.microsoft.com/fwlink/?linkid=525137 ->
https://officecdnmac.microsoft.com/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/Microsoft_Outlook_16.81.24011420_Installer.pkg
appNewVersion=16.81.24011420
```
If latest version is parsed from the MAU XML file (new behavior):
```
https://officecdn.microsoft.com/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/0409OPIM2019.xml ->
# updater
https://officecdnmac.microsoft.com/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/Microsoft_Outlook_16.81.24012814_Updater.pkg
appNewVersion=16.81.24012814

# delta updater
https://officecdnmac.microsoft.com/pr/C1297A47-86C4-4C1F-97FA-950631F94777/MacAutoupdate/Outlook_16.81.24011420_to_16.81.24012814_Delta.pkg
appNewVersion=16.81.24012814
```

In this case, when using current Installomator behavior:
If `appversion=16.81.24012814` (because it has already been updated by MAU or by other means) Installomator will keep trying to update it again and again because `appversion != appNewVersion`
OR
if `appversion=16.81.24011420` Installomator will report that it is current because `appversion == appNewVersion` even though the newest available version is actually 16.81.24012814

A true representation of an apps current version can be obtained by using a combination of the two data sources. This method has the additional benefit of grabbing the updater package URL (or delta update package URL) when an update is appropriate (based on the standard logic used by Installomator), possibly decreasing the download/install time.

I've also added logic that will check if MAU is available and functional before selecting `updateTool` as the update method. If MAU is unavailable, the update/install will continue using the downloadURL.

### DIFF
--- a/fragments/arguments.sh
+++ b/fragments/arguments.sh
@@ -99,6 +99,18 @@ if [[ ! -x $DIALOG_CMD ]]; then
     DIALOG_CMD_FILE=""
 fi
 
+# scot: moved here from main.sh to ensure all arguments are available before executing labels
+# MARK: finish reading the arguments:
+while [[ -n $1 ]]; do
+    if [[ $1 =~ ".*\=.*" ]]; then
+        # if an argument contains an = character, send it to eval
+        printlog "setting variable from argument $1" INFO
+        eval $1
+    fi
+    # shift to next argument
+    shift 1
+done
+
 # MARK: labels in case statement
 case $label in
 longversion)
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -638,7 +638,10 @@ installFromPKG() {
 
     deduplicatelogs "$spctlOut"
 
-    if [[ $spctlStatus -ne 0 ]] ; then
+# scot: Microsoft does not notarize their Delta update packages, added extra check to allow for this
+    if [[ $spctlStatus -eq 3 ]] && [[ $expectedTeamID == "UBF8T346G9" ]] && [[ $downloadURL =~ "Delta" ]]; then
+       printlog "Microsoft does not notarize Delta updaters...continuing."
+    elif [[ $spctlStatus -ne 0 ]] ; then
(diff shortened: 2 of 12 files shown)


scriptingosx:
see also #1499, #1519

Theile:
In order to handle changes to labels, we moved the reading of arguments to AFTER the label descriptions. I'm not so sure we want this moving back. We would rather add a function to run AFTER that labels and arguments have been read, but that might not be there before version 11.

And not so certain we will open for the un-notarized delta pkg, as we justify our scripted procedure for updates by being strict in this.

Can you somehow modify this PR to work only by changing labels, and then we can talk internally to see what we can do with the other changes? And you can think with the function mentioned above.

Theile:
I have made PR #1648 to read arguments before AND after the case statement with labels, so please remove this part here, thank you.

Theile:
I have removed the arguments reading changes in `arguments.sh` and `main.sh`.

scriptingosx:
please also see #1239 and #1455

please resolve merge conflicts before we will consider this

PR #1493, opened 2024-02-16
New Label: Apache NetBeans

jcmbowman:


kryptonit1:
The downloadURL does not work. It needs to be changed to `https://archive.apache.org/dist/netbeans/netbeans-installers/${appNewVersion}/Apache-NetBeans-${appNewVersion}.pkg`.

jcmbowman:
> The downloadURL does not work. It needs to be changed to `https://archive.apache.org/dist/netbeans/netbeans-installers/${appNewVersion}/Apache-NetBeans-${appNewVersion}.pkg`.

Change has been made.

scriptingosx:
Thank you!

PR #1514, opened 2024-02-26
Update mendeleyreferencemanager.sh

kryptonit1:
The current appNewVersion variable returns the latest version in the format "2.110.2-x64" instead of just "2.110.2". This change fixes that problem.


scriptingosx:
Thank you!

PR #1525, opened 2024-03-06
Update macadminspython.sh

sintichn:
Label was using old team id. I updated to the new macadmins org.


BigMacAdmin:
Duplicate of #1478

PR #1532, opened 2024-03-11
Update sync label for blockingProcess

fleish:
This ensures all Sync & syncworker processes are killed before updating

### DIFF
--- a/fragments/labels/sync.sh
+++ b/fragments/labels/sync.sh
@@ -4,4 +4,5 @@ sync)
     downloadURL="https://www.sync.com/download/apple/Sync.dmg"
     appNewVersion="$(curl -fs "https://www.sync.com/blog/category/desktop/feed/" | xpath '(//channel/item/title)[1]' 2>/dev/null | sed -E 's/^.* ([0-9.]*) .*$/\1/g')"
     expectedTeamID="7QR39CMJ3W"
+    blockingProcesses=( "Sync" "sync-worker.exe" )
     ;;


BigMacAdmin:
I've noticed while testing this PR that `appNewVersion` is not pulling the correct installation version.

@fleish Since this is not relevant to your PR, I will accept your change, but if you use this label I suggest you look at improving the `appNewVersion` code to make it more functional.

BigMacAdmin:
Thank you!

fleish:
> I've noticed while testing this PR that `appNewVersion` is not pulling the correct installation version.
> 
> @fleish Since this is not relevant to your PR, I will accept your change, but if you use this label I suggest you look at improving the `appNewVersion` code to make it more functional.

Thanks for pointing this out, @BigMacAdmin. I guess it's lucky for me Installomator doesn't run for this on client machines unless JAMF identifies they need an update.

There are two weird things I noticed in my own testing, one of which I could use your insight on. When I run the command for appNewVersion explicitly on a Monterey machine it doesn't return any data due to a syntax error. Removing the `/dev/null` error redirect reveals it's related to `xpath`:

```curl  "https://www.sync.com/blog/category/desktop/feed/" | xpath '(//channel/item/title)[1]' | sed -E 's/^.* ([0-9.]*) .*$/\1/g'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Usage:
/usr/bin/xpath5.30 [options] -e query [-e query...] [filename...]

If no filenames are given, supply XML on STDIN. You must provide at
least one query. Each supplementary query is done in order, the
previous query giving the context of the next one.

Options:

-q quiet, only output the resulting PATH.
-s suffix, use suffix instead of linefeed.
-p postfix, use prefix instead of nothing.
-n Don't use an external DTD.
100 15910    0 15910    0     0  36490      0 --:--:-- --:--:-- --:--:-- 36490
curl: (23) Failed writing body (0 != 2890)```

 Adding a `-e` to the command seems to fix this, but oddly when the command runs automatically via Installomator I don't even see `xpath` referenced but it does return the latest value it finds on the desktop blog page:

```+./Installomator.sh:7082> downloadURL=https://www.sync.com/download/apple/Sync.dmg
+./Installomator.sh:7083> appNewVersion=+./Installomator.sh:7083> appNewVersion=+./Installomator.sh:
7083> curl -fs https://www.sync.com/blog/category/desktop/feed/
+./Installomator.sh:7083> appNewVersion=+./Installomator.sh:7083> sed -E 's/^.* ([0-9.]*) .*$/\1/g'
+./Installomator.sh:7083> appNewVersion=2.2.32
+./Installomator.sh:7084> expectedTeamID=7QR39CMJ3W
+./Installomator.sh:7085> blockingProcesses=( Sync sync-worker.exe )```

I suspect the `-e` piece has to do with the below function, but the fact I don't see `xpath` being referenced in the debug output using `zsh -x` is puzzling:

```# Handling of differences in xpath between Catalina and Big Sur
xpath() {
        # the xpath tool changes in Big Sur and now requires the `-e` option
        if [[ $(sw_vers -buildVersion) > "20A" ]]; then
                /usr/bin/xpath -e $@
                # alternative: switch to xmllint (which is not perl)
                #xmllint --xpath $@ -
        else
                /usr/bin/xpath $@
        fi
}```

In any event, the second issue is perhaps the one you are noticing and that is the "new" version published in the blog and picked up by Installomator does not match the actual latest version. This has long been an issue with Sync where their version numbers don't sync up between what they note in the blog and what they include in the applicaiton. I don't know why this is and can try reaching out to them about normalizing them to match but I wouldn't hold my breath they will do so given how they've responded historically.

Thanks for calling this out and for merging the PR so I'll be able to update to the next version of Installomator without having to worry about losing my manual addition of this fix.

PR #1534, opened 2024-03-12
Create appiuminspector.sh

Theile:
Apium Desktop is unsecure and should not be included.


BigMacAdmin:
@Theile This label is not working because the downloaded .app is not notarized. Please reach out to the developer, and/or close this PR.

PR #1296, opened 2023-10-30
Update Xerox Labels' Download URLs for macOS 14

rougegoat:
xeroxphaser7800 and xeroxworkcentre7800 were both set to look at the macOS 11 URLs, which stopped getting updated at version 5.13.2.  Updated them to look at the macOS 14 ones so they get the current build.

PR #1554, opened 2024-03-20
Update Download URLs for Xerox Printer Drivers

rougegoat:
Resubmitting with only the fragments edited as requested.  See Pull Request #1296 for original submission


BigMacAdmin:
Thank you!

PR #1556, opened 2024-03-21
Fix Version GoogleChromePKG

AymenAbb:
Agreed upon solution requested by @BigMacAdmin as provided by @fraserhess and @mjabroux from #1366. Didn't see anyone else doing a PR on this and it's been a bit.


KUSU2:
This fix deletes the download line instead of replacing the new version line (it now has two different new version lines).  May want to pull it and re-submit.

AymenAbb:
butterfingers... thank you

PR #1550, opened 2024-03-19
Slack changed where the version is. Updated the appNewLabel variable

karsondude97:
```
2024-03-18 16:34:24 : REQ   : slack : ################## Start Installomator v. 10.6beta, date 2024-03-18
2024-03-18 16:34:24 : INFO  : slack : ################## Version: 10.6beta
2024-03-18 16:34:24 : INFO  : slack : ################## Date: 2024-03-18
2024-03-18 16:34:24 : INFO  : slack : ################## slack
2024-03-18 16:34:24 : DEBUG : slack : DEBUG mode 1 enabled.
2024-03-18 16:34:24 : INFO  : slack : SwiftDialog is not installed, clear cmd file var
2024-03-18 16:34:25 : INFO  : slack : setting variable from argument DEBUG=0
2024-03-18 16:34:25 : DEBUG : slack : name=Slack
2024-03-18 16:34:25 : DEBUG : slack : appName=
2024-03-18 16:34:25 : DEBUG : slack : type=dmg
2024-03-18 16:34:25 : DEBUG : slack : archiveName=
2024-03-18 16:34:25 : DEBUG : slack : downloadURL=https://slack.com/ssb/download-osx-silicon
2024-03-18 16:34:25 : DEBUG : slack : curlOptions=
2024-03-18 16:34:25 : DEBUG : slack : appNewVersion=4.37.94
2024-03-18 16:34:25 : DEBUG : slack : appCustomVersion function: Not defined
2024-03-18 16:34:25 : DEBUG : slack : versionKey=CFBundleShortVersionString
2024-03-18 16:34:25 : DEBUG : slack : packageID=
2024-03-18 16:34:25 : DEBUG : slack : pkgName=
2024-03-18 16:34:25 : DEBUG : slack : choiceChangesXML=
2024-03-18 16:34:25 : DEBUG : slack : expectedTeamID=BQR82RBBHL
2024-03-18 16:34:25 : DEBUG : slack : blockingProcesses=
2024-03-18 16:34:25 : DEBUG : slack : installerTool=
2024-03-18 16:34:25 : DEBUG : slack : CLIInstaller=
2024-03-18 16:34:25 : DEBUG : slack : CLIArguments=
2024-03-18 16:34:25 : DEBUG : slack : updateTool=
2024-03-18 16:34:25 : DEBUG : slack : updateToolArguments=
2024-03-18 16:34:25 : DEBUG : slack : updateToolRunAsCurrentUser=
[...]

PR #1566, opened 2024-03-25
Update slack.sh

theadamcraig:
the appNewVersion was reporting "x64" for intel and "arm64" for Arm. Adjusting the field from 6 to 7 resulted in it properly pulling the new version so that installomator will no longer attempt unnecessary installs


theadamcraig:
This is a duplicate of https://github.com/Installomator/Installomator/pull/1550

Issue #1543, opened 2024-03-14
Google Chomre - new version detection not working


tbcahill:
The version detection for label googlechromepkg is not working.
When checking if an update is needed the output is:
   " INFO  : googlechromepkg : Latest version not specified."

This results in the installer being downloaded & installed even when the latest version is already installed.

Full output: 
2024-03-12 11:29:03 : REQ   : googlechromepkg : ################## Start Installomator v. 10.5, d
ate 2023-10-15
2024-03-12 11:29:03 : INFO  : googlechromepkg : ################## Version: 10.5
2024-03-12 11:29:03 : INFO  : googlechromepkg : ################## Date: 2023-10-15
2024-03-12 11:29:03 : INFO  : googlechromepkg : ################## googlechromepkg
2024-03-12 11:29:03 : INFO  : googlechromepkg : SwiftDialog is not installed, clear cmd file var
2024-03-12 11:29:03 : INFO  : googlechromepkg : setting variable from argument NOTIFY=success
2024-03-12 11:29:03 : INFO  : googlechromepkg : setting variable from argument BLOCKING_PROCESS_A
CTION=prompt_user_then_kill
2024-03-12 11:29:03 : INFO  : googlechromepkg : setting variable from argument IGNORE_APP_STORE_A
PPS=yes
2024-03-12 11:29:03 : INFO  : googlechromepkg : setting variable from argument REOPEN=yes
2024-03-12 11:29:03 : INFO  : googlechromepkg : BLOCKING_PROCESS_ACTION=prompt_user_then_kill
2024-03-12 11:29:03 : INFO  : googlechromepkg : NOTIFY=success
2024-03-12 11:29:03 : INFO  : googlechromepkg : LOGGING=INFO
2024-03-12 11:29:03 : INFO  : googlechromepkg : LOGO=/System/Applications/App Store.app/Contents/
Resources/AppIcon.icns
2024-03-12 11:29:03 : INFO  : googlechromepkg : Label type: pkg
2024-03-12 11:29:03 : INFO  : googlechromepkg : archiveName: Google Chrome.pkg
2024-03-12 11:29:03 : INFO  : googlechromepkg : no blocking processes defined, using Google Chrom
e as default
2024-03-12 11:29:03 : INFO  : googlechromepkg : App(s) found: /Applications/Google Chrome.app
2024-03-12 11:29:03 : INFO  : googlechromepkg : found app at /Applications/Google Chrome.app, ver
[...]

PR #1569, opened 2024-03-26
Update googlechromepkg.sh

AymenAbb:
Agreed upon solution requested by @BigMacAdmin as provided by @fraserhess and @mjabroux from #1366. Didn't see anyone else doing a PR on this and it's been a bit. Second commit due to my mistake on #1556


scriptingosx:
see also #1366, #1367, #1354, #1543

scriptingosx:
Thank you!

AymenAbb:
Yes we can, their chrome versions seem to be in sync. Will pr when I can.

PR #1374, opened 2023-12-14
ZuluJDK DownloadURL Fixes

drtaru:
ZuluJDK DMG Downloads were failing.
Fixed DownloadURL logic for all versions to create the correct functional URL.

PR #1570, opened 2024-03-27
Update zulujdk8.sh

kryptonit1:
Fixes the downloadURL variable for the zulujdk8 label which was not fully addressed by the original author in #1374.


BigMacAdmin:
Thank you!

PR #1571, opened 2024-03-27
Create foxitreader

jshtux:
label for installing foxit reader


jshtux:
had to fix some things

PR #1576, opened 2024-03-31
Update functions.sh - Stop overriding newer versions

aduffner:
It must be checked whether there are interferences with version numbers, in the hope that these are always incremental, so there should be no problems.


scriptingosx:
also duplicate of #2133

(I prefer the solution there since it uses `is-at-least` rather than shelling out to `sort -V`)

PR #1591, opened 2024-04-05
Update googlechromeenterprise.sh

AymenAbb:
Same implementation as #1569. I have not tested this yet but the official enterprise version for MacOS syncs up with the regular version so I'm assuming this is proper. Should fix #1367 and #1354.


BigMacAdmin:
Thank you!

PR #1595, opened 2024-04-10
Add blockingProcesses to garminexpress

pmex:
Apparently Garmin Express also has a Service process, which I forgot in min initial PR.
Installing the pkg with Installomator doesn't seem to quit the service process, so it needs to be quit before installation.


BigMacAdmin:
Thank you!

PR #1623, opened 2024-04-19
Fix Docker label

KMTMark:
With the 4.29 version the static url used as the download URL wasn't pointing to the correct version so was retrieving 4.28.
Using RSS feed to extract downloadurl variable with this revision.


acodega:
This ended up being a duplicate PR.

PR #1627, opened 2024-04-24
Update eclipsetemurin17.sh

kryptonit1:
Added logic for determining architecture (Apple Silicon or Intel).

Also removed the packageID variable since the appCustomVersion function is used to determine installed version.


BigMacAdmin:
Thank you!

PR #1638, opened 2024-04-30
Update clickshare.sh

wakco:
Read the download url json with getJSONValue, and get a more up to date app version.


scriptingosx:
Thank you!

PR #1659, opened 2024-05-18
Update filezilla.sh

reuven-cohen:
Updated CURL to include -A function due to page not populating correctly without it.


scriptingosx:
Thank you!

PR #1744, opened 2024-07-10
updated zulu to no longer generate invalid package urls

cvgs:
The current fragment started generating wrong download URLs like the one below (see the leading "19"):
https://cdn.azul.com/zulu/bin/19-ca-jdk17.0.11-macosx_aarch64.dmg

- downloadURL logic is now in one regular expression and more specific
- replace + with . in appCustomVersion so that it can be properly compared
- unified version logic for all fragments to the clean CFBundleIdentifier method
- removed comment from appNewVersion (because it works now)

### DIFF
--- a/fragments/labels/zulujdk11.sh
+++ b/fragments/labels/zulujdk11.sh
@@ -3,11 +3,11 @@ zulujdk11)
     type="pkgInDmg"
     packageID="com.azulsystems.zulu.11"
     if [[ $(arch) == i386 ]]; then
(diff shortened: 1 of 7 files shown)


cvgs:
The second commit further simplifies the label by removing appCustomVersion and switching to extracting appVersion from the package ID. To allow version comparison, we now change the last period in appNewVersion to a plus sign. (like appNewVersion=17.50+19).
This only works in combination with [this other PR](https://github.com/Installomator/Installomator/pull/1750), which fixes package version extraction.

PR #1762, opened 2024-07-26
Update Cloudflare WARP download link and install type

leifahlgrimm-ada:
Cloudflare seems to have changed their download URL for WARP, where the old URL now returns a 500 error. This adaptation changes the URL and downloads the pkg directly.


rderewianko:
This issue predates #1778 however, that one has had a bunch of comments with testing.

PR #1620, opened 2024-04-18
Update opera.sh

kryptonit1:
The downloadURL variable in the current label is not working (#1619). This PR fixes that problem.

-----

2024-04-18 15:38:09 : REQ   : opera : ################## Start Installomator v. 10.6beta, date 2024-04-18
2024-04-18 15:38:09 : INFO  : opera : ################## Version: 10.6beta
2024-04-18 15:38:09 : INFO  : opera : ################## Date: 2024-04-18
2024-04-18 15:38:09 : INFO  : opera : ################## opera
2024-04-18 15:38:09 : INFO  : opera : SwiftDialog is not installed, clear cmd file var
2024-04-18 15:38:11 : INFO  : opera : BLOCKING_PROCESS_ACTION=tell_user
2024-04-18 15:38:11 : INFO  : opera : NOTIFY=success
2024-04-18 15:38:11 : INFO  : opera : LOGGING=INFO
2024-04-18 15:38:11 : INFO  : opera : LOGO=/System/Applications/App Store.app/Contents/Resources/AppIcon.icns
2024-04-18 15:38:11 : INFO  : opera : Label type: dmg
2024-04-18 15:38:11 : INFO  : opera : archiveName: Opera.dmg
2024-04-18 15:38:11 : INFO  : opera : no blocking processes defined, using Opera as default
2024-04-18 15:38:11 : INFO  : opera : name: Opera, appName: Opera.app
2024-04-18 15:38:11.227 mdfind[9772:9299006] [UserQueryParser] Loading keywords and predicates for locale "en_US"
2024-04-18 15:38:11.227 mdfind[9772:9299006] [UserQueryParser] Loading keywords and predicates for locale "en"
2024-04-18 15:38:11.291 mdfind[9772:9299006] Couldn't determine the mapping between prefab keywords and predicates.
2024-04-18 15:38:11 : WARN  : opera : No previous app found
2024-04-18 15:38:11 : WARN  : opera : could not find Opera.app
2024-04-18 15:38:11 : INFO  : opera : appversion:
2024-04-18 15:38:11 : INFO  : opera : Latest version of Opera is 109.0.5097.45
2024-04-18 15:38:11 : REQ   : opera : Downloading https://download3.operacdn.com/ftp/pub/opera/desktop/109.0.5097.45/mac/Opera_109.0.5097.45_Setup.dmg to Opera.dmg
2024-04-18 15:38:13 : REQ   : opera : no more blocking processes, continue with update
[...]

PR #1774, opened 2024-08-02
Updating Opera Fragment to use new method for detecting installURL

kbjoin:
Changing the detection method used to obtain the downloadURL variable.


acodega:
Duplicate #1620

PR #1609, opened 2024-04-16
Select pgAdmin download based on architecture

fraserhess:
Output:
```
% utils/assemble.sh pgadmin4
2024-04-16 08:43:22 : REQ   : pgadmin4 : ################## Start Installomator v. 10.6beta, date 2024-04-16
2024-04-16 08:43:22 : INFO  : pgadmin4 : ################## Version: 10.6beta
2024-04-16 08:43:22 : INFO  : pgadmin4 : ################## Date: 2024-04-16
2024-04-16 08:43:22 : INFO  : pgadmin4 : ################## pgadmin4
2024-04-16 08:43:22 : DEBUG : pgadmin4 : DEBUG mode 1 enabled.
2024-04-16 08:43:23 : DEBUG : pgadmin4 : name=pgAdmin 4
2024-04-16 08:43:23 : DEBUG : pgadmin4 : appName=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : type=dmg
2024-04-16 08:43:23 : DEBUG : pgadmin4 : archiveName=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : downloadURL=https://ftp.postgresql.org/pub/pgadmin/pgadmin4/v8.5/macos/pgadmin4-8.5-arm64.dmg
2024-04-16 08:43:23 : DEBUG : pgadmin4 : curlOptions=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : appNewVersion=8.5
2024-04-16 08:43:23 : DEBUG : pgadmin4 : appCustomVersion function: Not defined
2024-04-16 08:43:23 : DEBUG : pgadmin4 : versionKey=CFBundleShortVersionString
2024-04-16 08:43:23 : DEBUG : pgadmin4 : packageID=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : pkgName=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : choiceChangesXML=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : expectedTeamID=TCHGL2R7C5
2024-04-16 08:43:23 : DEBUG : pgadmin4 : blockingProcesses=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : installerTool=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : CLIInstaller=
2024-04-16 08:43:23 : DEBUG : pgadmin4 : CLIArguments=
[...]

PR #1796, opened 2024-08-22
Update pgadmin4.sh to support arm64 arch

PatrickRoumanoff:
The current `pgadmin4.sh` label default to `x86_64` when a `arm64` version is also available.

I also introduces those other changes:
* downloadParent url updated to use the same between version fetching and dmg fetching
* update regex to use `\d` instead of `[0-9]`
* update `sort` param from `-V` to `--sort-version` to be more explicit

This is my first PR - please let me know if this is the correct way to fix this issue.

Tested on my local (arm64)

### DIFF
--- a/fragments/labels/pgadmin4.sh
+++ b/fragments/labels/pgadmin4.sh
@@ -1,8 +1,13 @@
 pgadmin4)
     name="pgAdmin 4"
     type="dmg"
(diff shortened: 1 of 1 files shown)


PatrickRoumanoff:
this was a duplicate of https://github.com/Installomator/Installomator/pull/1609

PR #1853, opened 2024-09-04
Update crashplansmb.sh

orrison-CrashPlan:
CrashPlan SMB specific download URLs were removed. Updating to point to the default latest application link.


scriptingosx:
the previous urls in the existing labels also work, can you explain the change?

scriptingosx:
but from what I can tell, the CrashPlan and CrashPlan SMB download links are the same now? If that is correct, we should merge the labels?

orrison-CrashPlan:
Hello! That is correct, I'm fine with the merge, that would be a better plan than keeping them seperate.

PR #1862, opened 2024-09-06
Updated fragments for Google Drive and Microsoft Teams 2

tully-systima:
Google Drive app name that is pulled from existing fragment downloadURL is now named "Google Drive",  no longer "Google Drive File Stream", updated name variable to match 

Microsoft Teams app name that is pulled from existing fragment downloadURL is now named "Microsoft Teams", no longer "Microsoft Teams (work or school)", updated name variable to match 
Also added function to reload the Launch Agent for msupdate tool which helps with reliability and prevents some errors

### DIFF
--- a/fragments/labels/googledrive.sh
+++ b/fragments/labels/googledrive.sh
@@ -1,7 +1,7 @@
 googledrive|\
 googledrivefilestream)
     # credit: Isaac Ordonez, Mann consulting (@mannconsulting)
-    name="Google Drive File Stream"
+    name="Google Drive"
     type="pkgInDmg"
     if [[ $(arch) == "arm64" ]]; then
        packageID="com.google.drivefs.arm64"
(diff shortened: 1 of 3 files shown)


acodega:
Reposting from our Slack convo, for visibility.

Thanks! Couple things going on..
Label PRs really need to be submitted separately. Some exceptions are when the PR is for the same app, but the app label comes in different editions, like the zulu labels for example.

The Google Drive label name should be updated if the name is different but just FYI it's not a big deal since appName="Google Drive.app" is specified on the label. That overrides Installomator using the name as the app name to look for.

In the Teams label, how does changing the app name affect systems when "Microsoft Teams (work or school)" is installed and hasn't been updated to the new app? I know Microsoft is making this whole thing a can of worms. Also, you never need sudo in a script, Installomator runs as root. I know the bootout commands are new (right?) in Sonoma? What happens when they run on older macOS like Ventura? Could you compare your changes with #1239 and see if theres anything you can contribute on that PR perhaps?

PR #1942, opened 2024-10-04
Updating RDM Download URL

mechman27:
Updated RDM Free Download URL for label's .sh file.


mechman27:
Found that original URL was working as expected.

PR #1979, opened 2024-10-24
Update googlechrome.sh

illudium:
The googlechrome) recipe is still set to use omahaproxy.appspot.com (including the current latest, v 10.6) - but that site has been shut down, and refers us to https://chromiumdash.appspot.com

This updated recipe uses getJSONValue to parse the value of https://chromiumdash.appspot.com/fetch_releases?platform=Mac&channel=Stable&num=1 for the version number.


scriptingosx:
Thank you!

PR #2039, opened 2024-11-27
Update privileges.sh

MacMike077:
Installer is changed from .zip to .pkg

### DIFF
--- a/fragments/labels/privileges.sh
+++ b/fragments/labels/privileges.sh
@@ -1,7 +1,7 @@
 privileges)
     # credit: Erik Stam (@erikstam)
     name="Privileges"
-    type="zip"
+    type="pkg"
     downloadURL=$(downloadURLFromGit sap macOS-enterprise-privileges )
     appNewVersion=$(versionFromGit sap macOS-enterprise-privileges )
     expectedTeamID="7R5ZEU67FQ"


acodega:
We were talking about this on Slack.. I personally prefer we make a new label, `privileges2`. The apps fundamentally work differently, in the sense of how they are configured and what launchdaemons are needed. I'm not even sure if the config profile from v1 as-is will work with v2. It's the type of "update" you'd want to plan for and announcement before transitioning. I'm actually going through this at my current organization. Folks will still need to use a privileges 1 label to update their fleet, if applicable to latest and last version before 2.

I'd like to hear some opinions.

wakco:
I'd have to agree with @acodega, create a fresh label rather than modifying the existing one.

MacMike077:
More or less agree, and I feel the pain. But … The main function of installomator is providing the latest/greatest application version. With that in mind my 2 cents.

We cannot be responsible for a decision a developper makes I guess? I think we have see this in the past that, by the nature of how installomator works, you are forged to adopt a new version if the program fundamentally changes without a broken label.

Everybody with the need of Privileges is (should be) acting right now: Disable Installomator and providing a self made PKG to their fleet if they have the need for v1.x, or implementing v2 and move on.

If we make 2 labels, when are we getting rid of the original label? Or infusing it into one label (alias)? At that point we'll have the same discussion.

wakco:
There is precedent to consider, there are several other apps with multiple major version based labels in Installomator, so in this instance, the idea of another label for the version upgrade is consistent with existing Installomator labels.

acodega:
We've done this with other apps when it fundamentally changes in a new version. 1Password is an example. An app like Slack keeps on rolling from version 3 to 4. You don't need to *change* how you deploy Slack. You do with Privileges.

I recommend that the Privileges label should be changed so it doesn't grab `latest`, and this label should become Privileges 2. Or, we remove Privileges (since administrators can download 1.5.4 since it probably won't be updated further) and have a Privileges 2 label.

acodega:
I'm going to close this but please feel free to reopen it if needed or open a new PR for `privileges2`

PR #2053, opened 2024-12-02
`mdfind` commands fixed

Theile:
We have to specify the `kind` in the query more precisely to make sure it only finds applications. And now it's not `Application`, but `App`.

This will find apps only:
```
mdfind -name "$name" 'kMDItemKind:"App"'
```

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -264,11 +264,10 @@ getAppVersion() {
 #            targetDir="/Applications/Utilities"
 #        fi
     else
-    #    applist=$(mdfind "kind:application $appName" -0 )
         printlog "name: $name, appName: $appName"
-        applist=$(mdfind "kind:application AND name:$name" -0 )
+        applist=$(mdfind -name "$name" 'kMDItemKind:"App"' -0)
+#        applist=$(mdfind -name "$appName" 'kMDItemKind:"App"' -0)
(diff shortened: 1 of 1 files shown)


Theile:
Updated with use of other fields that are not localized in both kind and name:
```
mdfind 'kMDItemContentType:com.apple.application AND kMDItemFSName:"'"$name"'"'
```

fragments/functions.sh
```diff
@@ -264,11 +264,11 @@ getAppVersion() {
 #            targetDir="/Applications/Utilities"
 #        fi
     else
-    #    applist=$(mdfind "kind:application $appName" -0 )
         printlog "name: $name, appName: $appName"
-        applist=$(mdfind "kind:application AND name:$name" -0 )
+        # mdfind now handling if kind is either Application or App
+        applist=$(mdfind 'kMDItemContentType:com.apple.application AND kMDItemFSName:"'"$name"'"' -0 2>/dev/null)
```
scriptingosx, line 269: not a fan of your quotations here. They're not wrong per se, but entirely unreadable. Maybe use escaped quotes instead:

```
applist=$(mdfind "kMDItemContentType:com.apple.application AND kMDItemFSName:\"$name\"" -0 2>/dev/null)
```

Theile: Updated!

PR #2080, opened 2024-12-17
dialog.sh: added  archiveName and created dialoglauncher.sh

marcelpogorzelski:
Added archiveName to dialog.sh and created an installer for dialog-launcher.


acodega:
Bart "fixed" the file name issue in his repo so this issue no longer needs addressing. Thanks though.

PR #2086, opened 2024-12-21
Fixed AmazonWorkspaces

kevinstrick:
appNewVersion was failing; Updated it to include the DSA key needed for authorization


acodega:
Seems like it's working now without the key, can you confirm?

PR #2101, opened 2025-01-03
Update depnotify.sh to use appNewVersion

Mac-Nerd:
PKG is not published on github, but release notes and latest source are. Using that for appNewVersion to avoid reinstalling over current release.

Also updated URL to Jamf site, since nomad.menu was less reliable while testing.


acodega:
I think this is a duplicate of 2084

PR #2102, opened 2025-01-06
`RETURN_LABEL_NEW_VERSION` variable added

robjschroeder:
`RETURN_LABEL_NEW_VERSION` added to Installomator.sh. This allows Installomator to be ran with an app label to return only the latest available version of that app.

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -139,6 +139,11 @@ REOPEN="yes"
 #            use DEPNotify, then DEPNotify can present a more nice name to the user,
 #            instead of just the label name.
 
+# Only let Installomator return the appNewVerison of the label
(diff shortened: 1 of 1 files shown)


scriptingosx:
ok, I like this idea, but not the execution.

Is setting a variable to change the output behavior really the best option here? Variables have been the standard way to influence Installomator's behavior.

This feels like something that should be an option (something like `--appNewVersion`, but adding option parsing to the argument parsing sounds like a major change. Might be worth it, if it sets Installomator up better for the future... 🤔 

Either way, at the very least, change the variable to `RETURN_APP_NEW_VERSION`

it returns the new version of the _app_ not the labels.

PR #2303, opened 2025-04-01
GitHub auth & labelless (download label at label install time) version - Proof of Concept

wakco:
---
Yes.

No, this is a proof of concept for comparison/viewing by the maintainers.

No, see above.

**Fixes included:**
1. No point in checking if versionKey exists, when wanting a default, set it in the headers as a default that simply gets overridden. Also moved some log settings and `$SESSION` to the headers as well.
2. in `downloadURLFromGit`, moved the existence check of `$archiveName` to be part of setting `$filetype`, I didn't see much point in setting `$filetype` if it wasn't going to be used, which in turn meant what was 11 lines of code became only 7 lines of 3 additional for setting `$filetype`, and 4 of the 11 original 11 remaining to get the `$downloadURL`. I also had noticed a problem with some labels setting `$archiveName` to pattern matches, especially for Github, which I have seen cause problems with attempting to download and install the labels app/package due to the pattern match then failing to match a filename that is the pattern match, so added a reset of `$archiveName` to what the GitHub filename actually is.
3. I noticed a variation of my previous version of this for command line processing was adopted, but noticed it excluded appCustomVersion since it is a function not a variable, so added an exception for it.
4. It has two case statements for processing `$1` (the `$label`), the first one filters out version and longversion, as neither have any reason to have any further processing enabled, so after the case statement, all the code that was between `label=$1` and the label processing case statement (as well as the first run of processing the command line), are now between the version case statement and the label processing case statement.
5. Also move the "START" `printlog` entries to between the version case statement and the first round of command line argument processing.

**Additions:**
1. New `GITHUBAPI` setting, allows users to provide an appropriately permission'ed (i.e. read only) Github API token, to enable access to a larger 5000 hits per hour, instead of being limited to GitHub's public 60 hits per hour, useful for lab environments when behind a NAT'ed firewall.
2. New `checkRATEfromGit` function to log GitHub's rate limit status when `GITHUBAPI` is used, and confirm the token has the access required.
3. If a label doesn't exist within the script, it will now check for its existence on GitHub, download it and process it. The code here may still have an issue or two, such as a comment attached to the end of a line (instead of being on a line of it's own) is difficult to filter out (mostly fixed, insofar as all labels with comments on end of lines are now being processed properly), and may cause issues, as eval doesn't process multi-line commands, so while using ` ; ` as a separator (instead of `\n`) works, ~doesn't~ might not override a comment if it sneaks in (all current labels with comments on end of lines have the comments removed correctly to allow the label to be executed).
4. With 3. above, added an option (`--nolabels`) to `assemble.sh` to exclude adding labels in the script, allowing for the 11 thousand plus line script to be reduced by about 9 thousand lines to a little over 2 thousand lines.

Again, proof of concept, so allowing the maintainers to have a play to see what they think. Also I haven't rebuilt Installomator.sh in this branch, so testing will require rebuilding first, but also to limit the changes that are shown to the base fragments.

As this is a proof of concept, if there are any, I haven't checked. But like #2308, it fixes #2307, but in a different way that also fixes `longversion`.


wakco:
The `golang` label, is the first technically incompatible label, as it has a comment at the end of the line setting `appNewVersion`, however as that is not a required variable, the label does still work.

PR #2668, opened 2025-12-01
Fix dockutil check before adding to Dock

5h4rk1337:
dockutil runs now only if the app is really installed.

---
YES

NO

NO - Used the GitHub Web Editor.


uurazzle (changes requested):
HI @5h4rk1337 

Thanks for the contribution and for improving the dockutil handling logic.

That said, Installomator’s primary purpose is to install and update applications to their latest versions. While dockutil may be used as part of post-install actions, broader management of dockutil itself—such as conditional execution logic tied to its presence or lifecycle—is generally considered outside the core scope of Installomator.

In most environments, dockutil is managed separately via MDM scripts or configuration workflows, where administrators have more control over when and how Dock modifications are applied.

Given that, we should be careful about adding additional logic that extends beyond Installomator’s intended responsibility. It may be more appropriate to handle this behavior in downstream management tooling rather than within Installomator itself.

PR #2752, opened 2026-01-29
Update Camtasia name to include year 2024

a-martin253:
This application is still called "Camtasia 2024" so the current label is broken since even though it downloads the right file, it errors out because with the name being "Camtasia", it is then looking for "Camtasia.app" rather than "Camtasia 2024.app".

The name change of dropping the year only starts with 2025 and on.

---
Yes
Yes
Yes

### DIFF
--- a/fragments/labels/camtasia2024.sh
+++ b/fragments/labels/camtasia2024.sh
@@ -1,5 +1,5 @@
 camtasia2024)
-    name="Camtasia"
+    name="Camtasia 2024"
     type="dmg"
     sparkleData=$(curl -fsL -H 'User-Agent: Camtasia/2024.0.0' 'https://www.techsmith.com/redirect.asp?target=sparkleappcast&product=camtasiamac&ver=2024.0.0&lang=enu&os=mac')
     appNewVersion=$(


uurazzle (changes requested):
Hi @a-martin253,

Thanks for the PR. 

### Installomator log missing or non-zero exit code

We need a valid Installomator log in this PR before we can continue review.

The checklist expects a successful run with exit code 0. If your run exits non-zero (for example 3), please explain why that code is expected for this scenario and reference the Installomator exit code documentation:

- https://github.com/Installomator/Installomator/wiki/Installomator-Exit-Codes

To generate the log, run Installomator from Terminal and capture the output. You can enable `DEBUG=1` if helpful, but please **do not** set the logging level to DEBUG (`LOGGING=DEBUG`) as that produces excessive output. See the [Logging Level docs](https://github.com/Installomator/Installomator/wiki/Configuration-and-Variables#logging-level) for details.

Then paste the full output into this PR wrapped in a fenced code block:

```
paste log here
```

See [GitHub's code block docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks) if you need formatting help. Thanks!

### Backslash continuations should be removed

We found one or more variable/command definitions split across multiple lines using trailing backslashes (`\`). In Installomator labels, we prefer avoiding backslash line continuations for readability and consistency.

Please rewrite assignments like this:

```bash
appNewVersion=$(curl -fsL "https://update.code.visualstudio.com/api/releases/stable" | tr ',' '\n' | head -1 | tr -d '["')
```

instead of:

```bash
appNewVersion=$(curl -fsL "https://update.code.visualstudio.com/api/releases/stable" \
    | tr ',' '\n' \
    | head -1 \
    | tr -d '["')
```
For example

```
appNewVersion=$(
	echo "$sparkleData" | \
	xmllint -xpath 'string(//*[local-name()="item"][last()]/*[local-name()="shortVersionString"]/text())' -
)
```

```
downloadURL=$(
	echo "$sparkleData" | \
	xmllint -xpath 'string(//*[local-name()="item"][last()]/enclosure/@url)' -
)

```

Please update and re-push. Thanks!

uurazzle:
HI @a-martin253:

Thanks for submitting this. 

I agree with the issue you identified: Camtasia 2024 still installs as `Camtasia 2024.app`, so `name="Camtasia"` causes Installomator to look for the wrong app bundle.

This has already been fixed on `main` in the merged `camtasia2024` label:

https://github.com/Installomator/Installomator/blob/main/fragments/labels/camtasia2024.sh

Current `main` now has:

```bash
name="Camtasia 2024"
```

So, I will be closing the PR, but if you want it reopened, please comment back.

PR #2811, opened 2026-03-19
Fix jamovi label: switch to dl.jamovi.org and add arm64/x64 arch support

huexley:
---

Yes

**Is this pull request creating or modifying a label in the fragments/labels folder, and not Installomator.sh itself?**Yes

Yes

Yes

The jamovi download page (jamovi.org/download.html) now renders its 
download links via JavaScript, making curl-based HTML parsing unreliable 
(returns empty string, causing downloadURL to fall back to the homepage).

This fix switches to the static directory index at https://dl.jamovi.org/ 
which is reliably parseable with curl. It also adds proper arm64/x64 
architecture detection since jamovi now distributes separate DMGs per arch.
Additionally fixes appNewVersion which referenced undefined $downloadPATH.

Script result: 2026-03-19 11:28:44 : REQ   :  : shifting arguments for Jamf
2026-03-19 11:28:44 : INFO  : jamovi : Total items in argumentsArray: 0
2026-03-19 11:28:44 : INFO  : jamovi : argumentsArray: 
2026-03-19 11:28:44 : REQ   : jamovi : ################## Start Installomator v. 10.8, date 2025-03-28
2026-03-19 11:28:46 : REQ   : jamovi : Downloading https://dl.jamovi.org/jamovi-2.6.46.0-macos-arm64.dmg to jamovi.dmg
2026-03-19 11:29:04 : REQ   : jamovi : Installing jamovi
2026-03-19 11:29:34 : REQ   : jamovi : Installed jamovi, version 2.6.46.0
2026-03-19 11:29:36 : REQ   : jamovi : All done!
2026-03-19 11:29:36 : REQ   : jamovi : ################## End Installomator, exit code 0 

n/a

### DIFF
--- a/fragments/labels/jamovi.sh
+++ b/fragments/labels/jamovi.sh
@@ -2,19 +2,20 @@ jamovi)
     name="jamovi"
     type="dmg"
     baseURL="https://dl.jamovi.org"
-    
     if [[ $(arch) == "arm64" ]]; then
-    archSuffix="arm64"
+        archSuffix="arm64"
     elif [[ $(arch) == "i386" ]]; then
-    archSuffix="x64"
+        archSuffix="x64"
     else
-    printlog "ERROR: Unsupported architecture $(arch) for jamovi label" ERROR
-    exit 1
+        printlog "ERROR: Unsupported architecture $(arch) for jamovi label" ERROR
+        exit 1
     fi
     if [[ -n $jamoviLatest ]]; then
-    downloadFile=$(curl -sf "${baseURL}/" | grep -o "jamovi-2\.[7-9][^\"]*-macos-${archSuffix}\.dmg" | sort -V | tail -1)
+        downloadFile=$(curl -sf "${baseURL}/" | grep -o "jamovi-2\.[7-9][^\"]*-macos-${archSuffix}\.dmg" | sort -V | tail -1)
     else
-    downloadFile=$(curl -sf "${baseURL}/" | grep -o "jamovi-2\.6[^\"]*-macos-${archSuffix}\.dmg" | sort -V | tail -1)
+        downloadFile=$(curl -sf "${baseURL}/" | grep -o "jamovi-2\.6[^\"]*-macos-${archSuffix}\.dmg" | sort -V | tail -1)
     fi
     downloadURL="${baseURL}/${downloadFile}"
     appNewVersion="$(echo "$downloadFile" | cut -d '-' -f 2)"
+    expectedTeamID="9NCBP559AB"
+    ;;
--- a/fragments/labels/jamovi.sh
+++ b/fragments/labels/jamovi.sh
@@ -1,27 +1,27 @@
 jamovi)
     name="jamovi"
     type="dmg"
-    baseURL="https://www.jamovi.org"
+    baseURL="https://dl.jamovi.org"
 
-    # Detect architecture for separate arm64/x64 DMGs
     if [[ "$(arch)" == "arm64" ]]; then
         archSuffix="arm64"
     else
         archSuffix="x64"
     fi
 
-    # solid (stable) is listed first, current (latest) is listed second
     if [[ -n $jamoviLatest ]]; then
-        downloadPath=$(curl -sf "${baseURL}/download.html" \
-            | grep -o "/downloads/jamovi-[^\"]*-macos-${archSuffix}\.dmg" \
-            | tail -1)
+        # current = 2.7.x series (odd minor)
+        downloadFile=$(curl -sf "${baseURL}/" \
+            | grep -o "jamovi-2\.[7-9][^\"]*-macos-${archSuffix}\.dmg" \
+            | sort -V | tail -1)
     else
-        downloadPath=$(curl -sf "${baseURL}/download.html" \
-            | grep -o "/downloads/jamovi-[^\"]*-macos-${archSuffix}\.dmg" \
-            | head -1)
+        # solid = 2.6.x series (even minor)
+        downloadFile=$(curl -sf "${baseURL}/" \
+            | grep -o "jamovi-2\.6[^\"]*-macos-${archSuffix}\.dmg" \
+            | sort -V | tail -1)
     fi
 
-    downloadURL="${baseURL}${downloadPath}"
-    appNewVersion="$(echo "$downloadPath" | cut -d '-' -f 2)"
+    downloadURL="${baseURL}/${downloadFile}"
+    appNewVersion="$(echo "$downloadFile" | cut -d '-' -f 2)"
     expectedTeamID="9NCBP559AB"
     ;;
--- a/fragments/labels/jamovi.sh
+++ b/fragments/labels/jamovi.sh
@@ -1,14 +1,27 @@
 jamovi)
     name="jamovi"
     type="dmg"
-    downloadURL="http://www.jamovi.org"
-    if [[ -n $jamoviLatest ]]; then
-        downloadURL="${downloadURL}$(curl -s "$downloadURL/download.html" | grep macos | grep "download-button" | head -1 | cut -d '"' -f 4)"
+    baseURL="https://www.jamovi.org"
+
+    # Detect architecture for separate arm64/x64 DMGs
+    if [[ "$(arch)" == "arm64" ]]; then
+        archSuffix="arm64"
     else
-        downloadURL="${downloadURL}$(curl -s "$downloadURL/download.html" | grep macos | grep "download-button" | tail -1 | cut -d '"' -f 4)"
+        archSuffix="x64"
     fi
-    # The above is a cheat, they list both the "Latest" version and the "Solid" version twice on the page, but in opposing order.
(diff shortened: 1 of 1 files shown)


uurazzle (changes requested):
Hi @huexley,

Thanks for the PR. 

### Backslash continuations should be removed

We found one or more variable/command definitions split across multiple lines using trailing backslashes (`\`). In Installomator labels, we prefer avoiding backslash line continuations for readability and consistency.

Please rewrite assignments like this:

```bash
appNewVersion=$(curl -fsL "https://update.code.visualstudio.com/api/releases/stable" | tr ',' '\n' | head -1 | tr -d '["')
```

instead of:

```bash
appNewVersion=$(curl -fsL "https://update.code.visualstudio.com/api/releases/stable" \
    | tr ',' '\n' \
    | head -1 \
    | tr -d '["')
```

### Commented lines should be removed

We found comment lines added in the submitted label (lines starting with `#`).

Installomator labels should not contain comments — no descriptions, URL references, or commented-out code. Keeping labels free of comments makes review easier and reduces long-term maintenance noise.

If useful context exists, please include it in the PR description instead.

### Architecture guard missing (Apple Silicon only)

The label appears to target Apple Silicon (arm64) only. Installomator has a built-in convention for handling architecture-specific downloads. Rather than leaving the label without a guard, please wrap the download URL in an architecture check:

```bash
    downloadURL="<arm64-url>"
else
    printlog "<AppName> is only compatible with Apple Silicon (arm64) Macs." ERROR
    cleanupAndExit 95 "<AppName> requires Apple Silicon" ERROR
fi
```

This prevents the label from being run on Intel Macs and gives a clear error message if someone tries.

Please update and re-push. Thanks!

huexley:
Thanks for the review! I've addressed all three points:

Replaced backslash continuations with single-line pipelines
Removed inline comments
Added explicit i386 branch + unsupported-architecture guard with printlog ERROR and exit 1

uurazzle (changes requested):
Thanks for the contribution! A few things to address before this can merge:

**`i386` is not a valid return value of `arch` on macOS**
On Intel Macs, `arch` returns `x86_64`, so the `elif` branch never matches and Intel users always hit the error/exit path. Change `i386` to `x86_64`.

**`$jamoviLatest` is an undefined variable**
This doesn't follow any Installomator convention and will always evaluate as empty, meaning the `else` branch (2.6.x) is hardcoded in practice. If a current/latest channel split is needed, the pattern is two separate labels (e.g. `jamovi` and `jamovilatest`). Otherwise, just always fetch latest.

**Version-pinned regex will break on major version bumps**
Both `jamovi-2\.[7-9]` and `jamovi-2\.6` will silently return nothing when 3.x ships. Using `jamovi-[0-9]` is future-proof and lets `sort -V | tail -1` do the right thing.

**`appNewVersion` should be set — this is required per CONTRIBUTING.md**
The wiki notes that `appNewVersion` is expected for all labels where it's possible, and the [Label Variables Reference](https://github.com/Installomator/Installomator/wiki/Label-Variables-Reference) states: _"We will question new PRs without this field, and require a reason for not including it."_ The version is extractable from the download URL, so there's no reason to omit it.

**Intermediate variables aren't needed**
`baseURL` and `downloadFile` are local intermediates that can be folded into single-line assignments for `downloadURL` and `appNewVersion`, keeping things consistent with how other labels are written. See the [Label Variables Reference](https://github.com/Installomator/Installomator/wiki/Label-Variables-Reference) for examples.

Suggested revision:

```bash
jamovi)
    name="jamovi"
    type="dmg"
        archSuffix="arm64"
    elif [[ $(arch) == "x86_64" ]]; then
        archSuffix="x64"
    fi
    downloadURL=$(curl -sf "https://dl.jamovi.org/" | grep -o "jamovi-[0-9][^\"]*-macos-${archSuffix}\.dmg" | sort -V | tail -1 | sed "s|^|https://dl.jamovi.org/|")
    appNewVersion=$(echo "$downloadURL" | cut -d'/' -f4 | cut -d'-' -f2)
    expectedTeamID="9NCBP559AB"
    ;;
```

Also please make sure your PR includes an install log as noted in [CONTRIBUTING.md](https://github.com/Installomator/Installomator/blob/dev/CONTRIBUTING.md).

huexley:
Hi @uurazzle, thanks for the detailed review. Since your review, jamovi's download infrastructure has changed significantly, which affects several of the suggestions. Sharing what I found so the reasoning is transparent:
dl.jamovi.org no longer exists. It returns NXDOMAIN on all resolvers (verified against the institution's DNS, plus 1.1.1.1 and 8.8.8.8). The directory-index approach the PR was built on — and that your suggested revision relies on — can no longer be reached. It was live back in March when the original log was captured, but jamovi has since dismantled it.
No scrapable directory or version endpoint remains. I checked the obvious candidates:

www.jamovi.org/downloads/ (listing) → 302 to /404.html
dl-cdn.jamovi.org/latest-macos-arm64.dmg → 404 (no "latest" alias)
www.jamovi.org/version.txt → returns the homepage HTML, not a version
download.html → download links are injected via JavaScript, so curl only sees an empty/static shell

This is why appNewVersion can't be derived programmatically anymore — there's no machine-readable source for the current macOS version.
GitHub tags are not a reliable fallback. The jamovi/jamovi tags top out at v2.7.30, while the site currently ships 2.7.32 for macOS. Deriving the version from GitHub would pin clients to a stale release, which is worse than a hardcoded value.
Downloads are hotlink-protected. Direct curl requests to the file URL return a 404 unless a Referer: https://www.jamovi.org/ header is sent. With it, the URL resolves to the real DMG (application/x-apple-diskimage, ~410 MB). This is handled via curlOptions.

uurazzle (commented):
Hi @huexley:

Thanks for putting the original jamovi label together. 

I would recommend checking out and testing this label revision.

```
jamovi)
    name="jamovi"
    type="dmg"
        archSuffix="arm64"
        archSuffix="x64"
    else
        printlog "ERROR: Unsupported architecture $(arch) for jamovi label" ERROR
        exit 1
    fi
    curlOptions=(-H "Referer: https://www.jamovi.org/download.html")
    downloadFile=$(curl -fsL -H "Referer: https://www.jamovi.org/" "https://www.jamovi.org/download.html" | grep -Eo "jamovi-[0-9]+(\.[0-9]+)+-macos-${archSuffix}\.dmg" | sort -V | tail -n 1)
    appNewVersion=$(echo "$downloadFile" | sed -E "s/^jamovi-([0-9]+(\.[0-9]+)+)-macos-${archSuffix}\.dmg$/\1/")
    downloadURL="https://www.jamovi.org/downloads/${downloadFile}"
    expectedTeamID="9NCBP559AB"
    ;;
```


The main reason I’d prefer this revised version is that the previous source host, https://dl.jamovi.org, is no longer reliable from testing. DNS resolution failed locally, while the current vendor-supported download path is https://www.jamovi.org/downloads/..., which redirects cleanly to https://dl-cdn.jamovi.org/.... Using that path keeps the label aligned with the download flow jamovi is currently publishing.

The updated logic also avoids pinning behavior around the older 2.6 family unless a custom variable is set. The official download page currently advertises the current 2.7.x builds for both Apple Silicon and Intel, so the revised label discovers the latest matching macOS DMG for the active architecture from the vendor page and builds the direct download URL from that filename.

One important detail is the Referer header. jamovi’s download page redirects to the homepage unless it is requested with the right Referer, so the new label uses Referer: https://www.jamovi.org/ when reading the page and passes a Referer for the DMG download as well. That makes the behavior reproducible instead of depending on a host or page response that may vary by request context.

I also verified the payload directly. The current Apple Silicon DMG mounts as a standard compressed disk image, contains jamovi.app, reports CFBundleShortVersionString as 2.7.32.0, and is notarized with Developer ID Team ID 9NCBP559AB. CFBundleVersion is not present, so the default Installomator version key is the correct fit here.

uurazzle:
Hi @huexley,

Thank you for submitting this PR and for your work on the label. I tested the current implementation, but it does not appear ready to merge.

I encountered several issues during verification:

- `https://dl.jamovi.org` did not resolve, so the current `baseURL` may not be reliable.
- `versionFromGit jamovi/jamovi` does not appear to return the current packaged desktop release. GitHub reports `2.7.30` as the latest release, while jamovi’s official release notes list newer versions, including `2.7.38`.
- The jamovi website’s current metadata references version `2.7.34.0`, but the corresponding `/downloads/jamovi-2.7.34.0-macos-*.dmg` URLs returned an HTML/404 response rather than a valid DMG.
- Because I could not download and mount a current DMG, I was unable to verify the payload, application bundle version, or Team ID against the proposed label.

Since this PR has remained unresolved for some time and there is currently no working download source, we will close it for now.

If the DMG provides `jamovi.app` as a standard drag-and-drop application for `/Applications`, a future version of the label should likely avoid using `packageID`. It should also use the application bundle version for detection unless further testing indicates otherwise.

If you identify a reliable download source and would like to continue working on the label, please comment on this PR and we can reopen it.

PR #2864, opened 2026-04-28
New GitHub running to assemble and open a PR

acodega:
Our assemble GitHub Action does not open a PR, it commits directly to main. 

That's bad practice.

This will open a PR instead.


acodega:
Ignore this PR.

PR #2903, opened 2026-05-02
fix(libreofficelanguagepack_intl): use reliable version detection from documentfoundation.org

wucm667:

Yes

Yes

Yes

**Additional context**

The previous `appNewVersion` approach scraped the LibreOffice download page HTML using `grep dl_version_number`, which breaks when the page structure changes. This PR switches to directly listing the stable directory at `download.documentfoundation.org` and extracting the latest version via regex — a more reliable approach.

Also removed all comment lines from the label as per review guidelines.

**Installomator log**

```code
(Label tested on macOS, confirmed version detection works with the new documentfoundation.org source)
```

### DIFF
--- a/fragments/labels/libreofficelanguagepack_intl.sh
+++ b/fragments/labels/libreofficelanguagepack_intl.sh
@@ -1,19 +1,12 @@
 libreofficelanguagepack_intl)
     name="LibreOffice Language Pack"
-    # appName="LibreOffice.app"
-    #
-    # Reads the primary language of the system and installs the appropriate language pack for the latest LibreOffice STABLE version
-    # There is no language pack for the US English language and no installation is required other than the LibreOffice software itself.
-    # Use in combination and after installing Libre Office (STABLE Version)
-    # This label requires user interaction to complete the installation
-    #
     type="dmg"
     packageID="org.libreoffice.script.langpack"
     userLanguage=$(runAsUser defaults read .GlobalPreferences AppleLanguages | head -2 | tail -1 | tr -dc "[:alnum:]\-")
     if [[ "$userLanguage" == "en-US" ]]; then
         cleanupAndExit 0 "No installation of a language pack is necessary for the US-English language."
     fi
-    appNewVersion="$(curl -Ls https://www.libreoffice.org/download/download-libreoffice/ | grep dl_version_number | head -n 1 | cut -d'>' -f3 | cut -d'<' -f1)"
+    appNewVersion="$(curl -s https://download.documentfoundation.org/libreoffice/stable/ | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | tail -1)"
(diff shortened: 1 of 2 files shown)


uurazzle (changes requested):
Hi @wucm667,

Thanks for the PR. We found comment lines added in the submitted label (lines starting with `#`).

### Remove label comments

Installomator labels should not contain comments — no descriptions, URL references, or commented-out code. Keeping labels free of comments makes review easier and reduces long-term maintenance noise.

If useful context exists, please include it in the PR description instead. Thanks!

Detected in label(s): libreofficelanguagepack_intl

gilburns (changes requested):
Changes to Installomator.sh should not be part of the PR. Please only change the label.sh file.

wucm667:
@uurazzle @gilburns Thanks for the reviews! All issues addressed:

- Removed all comment lines from the label
- Removed the Installomator.sh change, only fragments/labels/libreofficelanguagepack_intl.sh is modified
- Filled in the full PR template

Rebased on latest main. Let me know if anything else needs fixing!

uurazzle (commented):
Hi @wucm667,

Thanks for the PR. We noticed one or more label identifiers or fragment filenames contain characters outside the Installomator label naming convention.

### Use lowercase alphanumeric label names

Installomator labels are the primary identifiers for software recipes. The [Label Variables Reference](https://github.com/Installomator/Installomator/wiki/Label-Variables-Reference) describes labels as entries in Installomator's zsh `case` statement, beginning with the label name and ending with `;;`. The fragment file name should also match the label name and use the `.sh` extension.

This check is not about the display name in `name="MyApp"`; title case is fine there. It is checking the label identifier and the fragment filename, such as `myapp)` and `fragments/labels/myapp.sh`.

Please make sure the label follows these naming requirements:

- Use lowercase letters only.
- Numbers are allowed.
- Hyphens and underscores may be used when needed.
- Do not use uppercase letters, spaces, or other special characters.
- The fragment filename must exactly match the label name, with `.sh` appended.

For example:

```bash
myapp)
    name="MyApp"
    appName="MyApp.app"
    ;;
```

should be saved as:

```text
fragments/labels/myapp.sh
```

For this label, please rename it to remove the unsupported character pattern. For example:

```bash
libreofficelanguagepackintl)
    name="LibreOffice Language Pack"
    appName="MyApp.app"
    ;;
```

Please rename the label and matching fragment file, then re-push. Thanks!

Detected in label(s): libreofficelanguagepack_intl
Note: the label identifier inside the file (e.g. `libreofficelanguagepack_intl)`) may already be correct — only the fragment filename needs to be renamed.

wucm667:
@uurazzle @gilburns Thanks for the reviews! Renamed the label and file from `libreofficelanguagepack_intl` to `libreofficelanguagepackintl` (removed underscore). All naming convention issues should be addressed now.

uurazzle (approved):
Hi @wucm667,

Thanks for the updates on this Installomator label. I am approving this PR.

gilburns (approved):

PR #2914, opened 2026-05-05
Updated governance.md, readme.md with new names

acodega:
With Gil and Richard joining, need to update the following.

- Added names to Governance.md 
- Added names to readme.md
- Added names to header.sh so names are updated in the script

Honestly I'd like to take names out of header.sh so we don't need to update that as we add people.. opinions?

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,63 +1,63 @@
-# Governance
+## Governance
 
-This project is maintained by a small group of core maintainers, with support from community contributors and reviewers. The goal is to keep the project healthy, responsive, and welcoming while maintaining quality and stability.
+This project is maintained by a small group of **core maintainers**, with support from **community maintainers** and contributors. The goal is to keep the project healthy, responsive, and welcoming while maintaining quality and stability.
+
+---
 
 ## Roles
 
 ### Contributors
-
 Anyone can contribute by:
 
-* Opening issues
-* Submitting pull requests
-* Suggesting improvements
-* Reporting bugs
+- Opening issues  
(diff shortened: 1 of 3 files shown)


gilburns:
Unless someone else feels strongly, I'd side with taking the names out of the script header. Same as taking label comment credits out.

BigMacAdmin:
Definitely doesn't need to be in the script itself, but I think leaving the names in the README is a good idea.

acodega:
I'm going to undo changing the header for now, so we can get governance and readme published. Those are more important.

uurazzle:
Hi 

A few miscellaneous suggestions and feedback.

Governance doesn’t guide how labels should source versions/downloads.

Add a short hierarchy of preferred approaches:

- Vendor API / structured feed
- GitHub Releases
- Sparkle appcast
- Stable redirect
- HTML scraping (last resort)

Reduces long-term maintenance burden and fragile implementations.

Governance focuses on contribution/approval, not ongoing health.

Add a simple statement:
 - Maintenance is best-effort and community-driven
 - Some labels may become outdated
 - Contributors are encouraged (not required) to revisit labels they’ve worked on

clarify maintainer escalation expectations
Gap: Not explicit how to handle uncertain or risky PRs.

Encourage reviewers to:
 - loop in other maintainers when unsure
 - avoid merging high-risk changes without confidence

gilburns (approved):

acodega:
Governance is only meant to explain how the teams work and what the relationship is to new PRs, not anything about the project or writing a PR.

Issue #2029, opened 2024-11-22
Broken app label: 1password8


tt19697:
**What's wrong with the app label?**
Unable to complete update when 1Password for Safari is installed > when Safari is **opened**. Keeps looping `killing process 1Password`. Also, label is not getting the right version. [This PR would resolve that issue.](https://github.com/Installomator/Installomator/pull/2009)

When Safari is quit manually, Installomator continues with the installation successfully.

**Installomator version**
Version: 10.7beta

**Additional context**
In order to resolve, please include Safari.app as a process blocking app. Or maybe a way to see if 1Pas Safari Extension is installed, to only then quit Safari to not bother non-Safari 1Pass Extension users, but that would make things more complicated... Maybe something for Installomator v11, to have IF statements available? :D #wecanalldream

**Installomator log**
```
2024-11-20 12:04:45 : REQ   : 1password8 : ################## Start Installomator v. 10.7beta, date 2024-09-02
2024-11-20 12:04:45 : INFO  : 1password8 : ################## Version: 10.7beta
2024-11-20 12:04:45 : INFO  : 1password8 : ################## Date: 2024-09-02
2024-11-20 12:04:45 : INFO  : 1password8 : ################## 1password8
2024-11-20 12:04:47 : INFO  : 1password8 : setting variable from argument NOTIFY=silent
2024-11-20 12:04:47 : INFO  : 1password8 : setting variable from argument BLOCKING_PROCESS_ACTION=kill
2024-11-20 12:04:47 : INFO  : 1password8 : setting variable from argument DEBUG=0
2024-11-20 12:04:47 : INFO  : 1password8 : BLOCKING_PROCESS_ACTION=kill
2024-11-20 12:04:47 : INFO  : 1password8 : NOTIFY=silent
2024-11-20 12:04:47 : INFO  : 1password8 : LOGGING=INFO
2024-11-20 12:04:47 : INFO  : 1password8 : LOGO=/System/Applications/App Store.app/Contents/Resources/AppIcon.icns
2024-11-20 12:04:47 : INFO  : 1password8 : Label type: pkg
2024-11-20 12:04:47 : INFO  : 1password8 : archiveName: 1Password.pkg
2024-11-20 12:04:47 : INFO  : 1password8 : found packageID com.1password.1password installed, version 8.10.50
2024-11-20 12:04:47 : INFO  : 1password8 : appversion: 8.10.50
[...]

PR #3042, opened 2026-06-23
Add forcefulQuit label option; enable for 1password8

mactesting12:
## Problem

`1password8` updates abort with `exit 11` ("could not quit all processes") whenever 1Password is running. 1Password 8 is a menu-bar-resident Electron app that traps **both** the AppleScript `quit` event **and** the SIGTERM that `pkill` sends, so it keeps running after a normal quit (only its window closes). `checkRunningProcesses` can therefore never clear it — even `BLOCKING_PROCESS_ACTION=kill` loops, because that path only sends a catchable SIGTERM. See #2029.

There is no launchd KeepAlive respawning it; the process simply survives every catchable signal Installomator currently sends. Nothing in the quit path ever sends an uncatchable SIGKILL.

## Fix

Add an opt-in label variable `forcefulQuit="YES"`. When set, `checkRunningProcesses` tells the app to quit, waits a short grace period, then escalates to `pkill -9` (SIGKILL). A process is only counted as still-blocking if it survives SIGKILL, so apps that honor the polite quit within the grace window are never force-killed and behavior for all existing labels is unchanged.

Enabled on the `1password8` label. Documented in `header.sh` alongside the other label variables.

## Notes

- Fragments only, per CONTRIBUTING — `Installomator.sh` not edited.
- The same SIGTERM-trap affects other menu-bar Electron apps; they can opt in the same way.

### DIFF
--- a/fragments/functions.sh
+++ b/fragments/functions.sh
@@ -324,6 +324,31 @@ checkRunningProcesses() {
                 printlog "found blocking process $x"
                 appClosed=1
 
+                # forcefulQuit (label opt-in): some apps — notably
+                # menu-bar-resident Electron apps like 1Password 8 — trap
+                # both the AppleScript `quit` event and the SIGTERM that
+                # `pkill` sends, and keep running. A normal quit only
+                # closes their window; the background process survives, so
+                # the actions below can never clear it and we abort with
+                # exit 11. When the label sets forcefulQuit=YES, try a
+                # polite quit once, give it a short grace period, then
+                # escalate to an uncatchable SIGKILL. Only counts as still
+                # blocking if it survives even that.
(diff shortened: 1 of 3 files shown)


acodega:
Is it a problem if the app is running? Does the 1Password installer take care of it? It's possible blocking processes are not a concern for that label and I don't think there's another label this is relevant for.

Also we'd prefer comments are a minimum, just have the comment in the header, but the function fragment or in the label.

gilburns:
I'd suggest this is potentially an educational issue to train your user population know to quit the menubar app when prompted for 1Password updates? Or optionally you could also remove "1Password 8" from the blocking list and the PKG install will kill it automatically for you.

PR #3124, opened 2026-07-23
Update inkscape.sh

### LINKED ISSUE #3119: Custom user-agent for Installomator
**What's the bug?**<br>I am helping out with Inkscape system administration and I got a complaint that Inkscape downloads were not working for someone. After some digging, I fould some curl requests that are coming from Installomator. We could have skipped some steps if Installomator identified itself instead of using the default curl user-agent.

**Have you confirmed this Issue has not been reported already? (Check open Issues and Pull requests)** I looked and did not see any.

**Installomator version**<br>What version are you running? You must include the version number and date. (Ex: Installomator v. 10.7beta, date 2024-08-31) Installomator v. 10.9 and Present in the `main` branch.

**Additional context**<br>Add any other context about the problem here.

I would recommend something like this instead:

```
Mozilla/5.0 Installomator/10.9 (+https://github.com/Installomator/Installomator)
```

**Installomator log**<br>
At the bottom of this report, provide a log, either by running Installomator in Terminal and saving the output, or by downloading the result log from your MDM. **Do not enable [Debug logging level](https://github.com/Installomator/Installomator/wiki/Configuration-and-Variables#logging-level) and please format the log [using a code block](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks)!**

Partial server log corresponding to [inkscape.sh](https://github.com/Installomator/Installomator/blob/main/fragments/labels/inkscape.sh):

```
REDACTED - - [20/Jul/2026:16:20:14 +0000] "GET /release/inkscape-1.4.4/mac-os-x/dmg/dl/ HTTP/2.0" 200 26399 "-" "curl/8.7.1" "-"
REDACTED - - [20/Jul/2026:16:20:18 +0000] "GET /release/inkscape-1.4.4/ HTTP/2.0" 200 36658 "-" "curl/8.7.1" "-"
```

I would also recommend adding delays between requests and caching such that the same requests are not repeated.

Edit: typo

TechnologyClassroom:
---
**Have you confirmed this pull request is not a duplicate?** Yes.

**Is this pull request creating or modifying a label in the fragments/labels folder, and not Installomator.sh itself?** Yes.

**Did you use [our editorconfig file](https://github.com/Installomator/Installomator/wiki/Contributing-to-Installomator)?** Yes.

This is related to https://github.com/Installomator/Installomator/issues/3119 and adds a user-agent to Installomator requests for Inkscape instead of the default curl user-agent and slows down scrape requests.

```
I do not use macOS myself.
```

Fixes https://github.com/Installomator/Installomator/issues/3119 for Inkscape at least.

### DIFF
--- a/fragments/labels/inkscape.sh
+++ b/fragments/labels/inkscape.sh
@@ -2,11 +2,13 @@ inkscape)
     name="Inkscape"
     type="dmg"
     appCustomVersion() { /Applications/Inkscape.app/Contents/MacOS/inkscape --version | cut -d " " -f2 }
-    appNewVersion=$(curl -fsL https://inkscape.org/release/  | grep "<title>" | grep -o -e "[0-9.]*")
+    appNewVersion=$(curl -A 'Mozilla/5.0 Installomator (+https://github.com/Installomator/Installomator)' -fsL https://inkscape.org/release/  | grep "<title>" | grep -o -e "[0-9.]*")
+    sleep 4
     if [[ $(arch) == arm64 ]]; then
-        downloadURL="https://media.inkscape.org/dl/resources/file/$(curl -fsL https://inkscape.org/release/inkscape-{$appNewVersion}/mac-os-x/dmg-arm64/dl/ | grep -o -m1 "Inkscape-.*.dmg")"
+        downloadURL="https://media.inkscape.org/dl/resources/file/$(curl -A 'Mozilla/5.0 Installomator (+https://github.com/Installomator/Installomator)' -fsL https://inkscape.org/release/inkscape-{$appNewVersion}/mac-os-x/dmg-arm64/dl/ | grep -o -m1 "Inkscape-.*.dmg")"
     elif [[ $(arch) == i386 ]]; then
-        downloadURL="https://media.inkscape.org/dl/resources/file/$(curl -fsL https://inkscape.org/release/inkscape-{$appNewVersion}/mac-os-x/dmg/dl/ | grep -o -m1 "Inkscape-.*.dmg")"
+        downloadURL="https://media.inkscape.org/dl/resources/file/$(curl -A 'Mozilla/5.0 Installomator (+https://github.com/Installomator/Installomator)' -fsL https://inkscape.org/release/inkscape-{$appNewVersion}/mac-os-x/dmg/dl/ | grep -o -m1 "Inkscape-.*.dmg")"
     fi
+    sleep 4
     expectedTeamID="SW3D6BB6A6"
     ;;


gilburns (changes requested):
This should not be necessary:  

`    appCustomVersion() { /Applications/Inkscape.app/Contents/MacOS/inkscape --version | cut -d " " -f2 }
`  

The app bundle uses CFBundleShortVersionString for the version and that is the default key that Installomator will check.  

Also the two `sleep` command are not needed.

TechnologyClassroom:
The sleep commands would not be needed if there were slow downs built in. If a collection of machines all use Installomator behind NAT, the sleep commands are very helpful for the server side.

That situation is not hypothetical.

uurazzle (changes requested):
Hi @TechnologyClassroom:

Thanks for the PR!!!

Please remove the `sleep 4` commands and resubmit the PR.

acodega:
I think a simple `sleep` would simply delay the flood by 4 seconds rather than space out requests. While I understand the request I think anything around timing should be in the hands of the administrator. (The one using Installomator at their organization)

Any enhancements to add randomization would be outside of the scope of the label, and for the sake of the original request (adding headers, which I like the idea of!) I'd submit this label without the sleeps.

BigMacAdmin:
@TechnologyClassroom can you help us understand better the logic on the `sleep 4` ? Is this simply to separate the "flood" of "Check foir app new version" from the following flood of "Now download this installer"?

If so, and if the dev/host is saying this will help in this particular case, I'm inclined to say we should make an exception. We're getting great engagement here from the developer side, and I think it's worth considering as a one-off (and it doesn't mean we have to accept `sleep` commands in all labels going forward.)

I agree with adam that randomized sleep between when labels are ingested vs. when the actual download begins is outside the scope here, but it's an interesting idea that may be worth exploring.

While I agree with "let the admin decide" in this case the dev is telling us what decision to make to have Installomator work best with their infrastructure, and we should give it the appropriate consideration.

TechnologyClassroom:
As Installomator currently works, Installomator directs three requests nearly simultaneously within one or two seconds. Adding sleep commands would at least space them out. This would still not be following the the robots.txt file's `Crawl-Delay` of 86400 seconds, but it would help.

If an organization has one Installomator instance and then deploys them to all machines, there is not much of an issue with this sort of thing. If a collection of individuals all use separate Installomator instances behind one CGNAT address, Installomator is indistinguishable from a DoS attack.

uurazzle:
Thanks for the detailed explanation, @TechnologyClassroom — the CGNAT scenario makes sense and is a fair point: from the server's perspective, a bunch of independently-triggered Installomator runs behind one shared IP really can look indistinguishable from a flood, even though sleep 4 doesn't come close to satisfying the actual Crawl-Delay.

That said, I'd lean toward not hardcoding the sleep in the label itself. A few reasons:

- Once one label ships with a hardcoded delay "because the vendor asked," it's hard to say no to the next one, and delay tuning isn't something we can validate or reason about per-label — it's an infrastructure/deployment decision, not a packaging one.
- A flat 4-second sleep helps your specific CGNAT case but does nothing for orgs with hundreds of machines checking in near-simultaneously from different IPs — it doesn't actually solve the underlying problem, just one instance of it.
- If we want to support this properly, it'd be better as an opt-in mechanism — e.g. a variable admins can set (SLEEP=4 or similar) that defaults to 0, so the behavior stays under administrator control per @acodega's original point, but orgs who need it (like yours) can enable it without forking the label.

If you're open to it, I'd suggest resubmitting without the unconditional sleep 4 and instead adding an optional, admin-configurable delay variable that defaults to off. That gets you the spacing you need without us baking a fixed timing assumption into the label for everyone.

Installomator already has a general mechanism for exactly this — runtime variable overrides passed as arguments after the label name. Any variable set in a label can be overridden without editing the script itself: you can override variables when Installomator is run by providing additional arguments in the form VARIABLE=new_value, with the application label as the first argument followed by optional variable overrides, in any order. [GitHub](https://github.com/Installomator/Installomator/wiki/Configuration-and-Variables)

So the pattern would be:
1. In the label, define the variable with a safe default (off by default, per acodega's point):
```
SLEEP_BEFORE_DOWNLOAD=${SLEEP_BEFORE_DOWNLOAD:-0}
```

then use it right before the download step:
```
sleep "$SLEEP_BEFORE_DOWNLOAD"
```
2. An admin who wants the delay just passes it as an argument when calling Installomator:
```
sudo /usr/local/Installomator/Installomator.sh labelname SLEEP_BEFORE_DOWNLOAD=4
```
3. In Jamf Pro specifically, this maps cleanly to a script parameter. If Installomator is deployed as a policy running a wrapper/script, the admin adds SLEEP_BEFORE_DOWNLOAD=4 as one of the script parameter values (or appends it in the script's Installomator.sh $4 SLEEP_BEFORE_DOWNLOAD=4-style invocation), scoped only to the Macs/networks that need it — e.g. the CGNAT'd fleet — without affecting anyone else's deployment.
This is the same override mechanism BigMacAdmin's own blog describes for pinning app versions — any value in an Installomator label can be overridden at runtime by passing the desired value as an argument — so it's a well-established pattern in the project, not a new concept for this PR. [The BigMacAdmin Blog](https://bigmacadmin.wordpress.com/2024/08/12/pinning-application-versions-with-installomator/)

Net effect: the label ships with zero delay by default (matching "let the admin decide"), and TechnologyClassroom's org — or anyone else with a similar CGNAT/shared-IP setup — just adds one argument to their existing Installomator call.

TechnologyClassroom:
I get that this is a patch and not a fix, but this would be a sound default behavior for Installomator interacting with Inkscape. I believe all automation should have at least a 4 second delay between subsequent requests to the same domain.

If you want to update the pull request, you are free to do so, but I am not going to remove the sleep commands or default them to 0.

TechnologyClassroom:
With the scenario I mentioned previously, your suggestion would require 60 different individuals to add manual configuration overrides.

uurazzle:
Hi @TechnologyClassroom:

We have another PR open and will have your request for a 4 second throttle included when its merged.https://github.com/Installomator/Installomator/pull/3144

If nothing else needs to be addressed with this ticket is it ok if we close it and keep the other PR open until its addressed, reviewed and merged?

```
inkscape)
    name="Inkscape"
    type="dmg"
    curlOptions=( -A "Installomator/$VERSION (+https://github.com/Installomator/Installomator)" )
    inkscapeReleasesJSON=$(curl -fsL "${curlOptions[@]}" "https://media.inkscape.org/media/releases.json")
    appNewVersion=$(getJSONValue "$inkscapeReleasesJSON" '[0].releases.filter(release => release.released).sort((a, b) => a.released.localeCompare(b.released)).pop().version')
        inkscapePlatform="macOS : dmg (arm64)"
        inkscapePlatform="macOS : dmg (Intel)"
    fi
    downloadURL=$(getJSONValue "$inkscapeReleasesJSON" "[0].releases.filter(release => release.version == '$appNewVersion')[0].platforms.filter(platform => platform.name == '$inkscapePlatform')[0].download")
    SLEEP_BEFORE_DOWNLOAD=${SLEEP_BEFORE_DOWNLOAD:-4}
    sleep "$SLEEP_BEFORE_DOWNLOAD"
    expectedTeamID="SW3D6BB6A6"
    ;;
```

TechnologyClassroom:
Yes, the new method with scraping `https://media.inkscape.org/media/releases.json` is better than this method. This takes the load off of www.inkscape.org and moves it to Fastly CDN.

gilburns:
@TechnologyClassroom - If that other PR is hitting the CDN, then we probably don't need to include the 4 second throttle at all?

TechnologyClassroom:
Please, slow down.

PR #3109, opened 2026-07-17
ChatGPT: Point label at unified ChatGPT app

ReArmedHalo:
**Have you confirmed this pull request is not a duplicate?**

Yes.

**Is this pull request creating or modifying a label in the fragments/labels folder, and not Installomator.sh itself?**

Label - It modifies `fragments/labels/chatgpt.sh` only.

**Did you use [our editorconfig file](https://github.com/Installomator/Installomator/wiki/Contributing-to-Installomator)?**

Yes. Only the two URL string values changed; no reformatting.

**Additional context**

OpenAI merged Codex, ChatGPT, and ChatGPT Work into a single `ChatGPT.app`. The `sidekick` endpoints this label used now serve **"ChatGPT Classic"** - the appcast channel title is literally `ChatGPT Classic` and its enclosure is `ChatGPT_Classic.pkg` - so `downloadURL` and `appNewVersion` were pointing at the wrong app.

This points them at the unified app, which OpenAI now distributes from the `codex-app-prod` bucket:

- `downloadURL`: `https://persistent.oaistatic.com/codex-app-prod/Codex.dmg`
- `appNewVersion`: `https://persistent.oaistatic.com/codex-app-prod/appcast.xml`

`Codex.dmg` now contains `ChatGPT.app`, and `appNewVersion` parses to the same version reported for the installed app. `expectedTeamID` `2DC432GLL2` is unchanged and the app is still arm64-only, so the existing Intel guard is kept.

I would like to suggest the old `codex` label be removed entirely, but did not stage that in this PR.

**Installomator log**

```
2026-07-17 14:04:52 : INFO  : chatgpt : ################## chatgpt
2026-07-17 14:04:52 : DEBUG : chatgpt : DEBUG mode 1 enabled.
2026-07-17 14:04:52 : DEBUG : chatgpt : name=ChatGPT
2026-07-17 14:04:52 : DEBUG : chatgpt : type=dmg
2026-07-17 14:04:52 : DEBUG : chatgpt : downloadURL=https://persistent.oaistatic.com/codex-app-prod/Codex.dmg
2026-07-17 14:04:52 : DEBUG : chatgpt : appNewVersion=26.715.21425
2026-07-17 14:04:52 : DEBUG : chatgpt : versionKey=CFBundleShortVersionString
2026-07-17 14:04:52 : DEBUG : chatgpt : expectedTeamID=2DC432GLL2
2026-07-17 14:04:53 : INFO  : chatgpt : Label type: dmg
[...]

PR #3087, opened 2026-07-10
Fix codex.sh

dan-snelson:
All questions must be filled out or your Pull Request will be closed for lack of information. The first three questions should be answered `Yes` before submitting the pull request.
---
**Have you confirmed this pull request is not a duplicate?**

Yes

**Is this pull request creating or modifying a label in the fragments/labels folder, and not Installomator.sh itself?**

Yes

**Did you use [our editorconfig file](https://github.com/Installomator/Installomator/wiki/Contributing-to-Installomator)?**

Yes

**Additional context** Add any other context about the label or fix here.

The Codex release channel packages `ChatGPT.app` with bundle ID `com.openai.codex`, while public ChatGPT uses the same app name with bundle ID `com.openai.chat`. This change keeps Codex in `/Applications/ChatGPT.localized`, reads the version and matching ZIP URL from the same Sparkle feed, and relies on Installomator's built-in `targetDir`-aware version lookup.

**Installomator log** At the bottom of this pull request, provide a log of a label run by running Installomator in Terminal and saving the output. `DEBUG=1` can be enabled but **do not enable [Debug logging level](https://github.com/Installomator/Installomator/wiki/Configuration-and-Variables#logging-level) and please format the log [using a code block](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks)!**

Please identify any issues fixed by your pull request by including the issue number. (Example: "Fixes #XXXX")

No linked issue.

```text
2026-07-23 13:54:22 : INFO  : codex : setting variable from argument DEBUG=1
2026-07-23 13:54:22 : REQ   : codex : ################## Start Installomator v. 10.10beta, date 2026-07-23
2026-07-23 13:54:22 : INFO  : codex : ################## codex
2026-07-23 13:54:22 : DEBUG : codex : DEBUG mode 1 enabled.
2026-07-23 13:54:22 : DEBUG : codex : name=Codex
2026-07-23 13:54:22 : DEBUG : codex : appName=ChatGPT.app
2026-07-23 13:54:22 : DEBUG : codex : type=zip
2026-07-23 13:54:23 : DEBUG : codex : downloadURL=https://persistent.oaistatic.com/codex-app-prod/ChatGPT-darwin-arm64-26.715.72359.zip
2026-07-23 13:54:23 : DEBUG : codex : appNewVersion=26.715.72359
2026-07-23 13:54:23 : DEBUG : codex : appCustomVersion function: Not defined
[...]

PR #3151, opened 2026-08-04
Fix chatgpt label: repoint stale download URL, alias chatgptclassic, clean up legacy app on upgrade

### LINKED ISSUE #3132: Broken app label: [chatgpt] has changed to chatgptclassic
The “chatgpt” label only works up to version 1.2026.183. 
Starting with version 1.2026.184, you'll need a new label: “chatgptclassic.” 

chatgptclassic)
    name="ChatGPT Classic"
    type="dmg"
    downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
    appNewVersion=""
    expectedTeamID="2DC432GLL2"
        appCustomVersion(){
        infoPlist="/Applications/ChatGPT Classic.app/Contents/Info.plist"
        if [ -f "${infoPlist}" ];then
            defaults read "${infoPlist}" CFBundleShortVersionString
        fi
    }
    ;;

This label only works for apple silicon  💻

Hac3rsp4ce:

Yes. #3109 attempted a fix but pointed `chatgpt` at the unified/codex-app-prod bucket
(the wrong app) and was closed. This PR keeps `chatgpt` as "ChatGPT Classic" instead,
matching what @dan-snelson described in #3087: the old sidekick endpoint is now
permanently the Classic app, and the new unified app is a separate concern outside
this label.

Yes — only `fragments/labels/chatgpt.sh` is touched.

**Did you use our editorconfig file?**

Yes.

**Additional context**

OpenAI rebranded the ChatGPT desktop app to "ChatGPT Classic" around July 2026. This
broke the `chatgpt` label in two ways:

1. The `downloadURL` (`sidekick/public/ChatGPT_Desktop_public_latest.dmg`) is still
   live but frozen on a pre-1.2026.184 build. The appcast-based `appNewVersion` check
   was unaffected by the rebrand and still correctly reports the real latest version —
   so rather than silently reporting "up to date," the label was stuck comparing a
   correct "184 is latest" against an installed stale 183, meaning it would re-download
   the same stale dmg on every run without ever converging. Repointed `downloadURL` at
   the actively-updated classic dmg; left the version-check logic untouched since it
   was already correct.
2. On a machine that already has the pre-184 `ChatGPT.app` installed, simply fixing the
   URL isn't enough — Installomator's built-in already-installed detection keys off
   `/Applications/$appName` (now `ChatGPT Classic.app`) with an `mdfind` fallback keyed
   on `$name`, neither of which recognizes `ChatGPT.app` as this app's predecessor. That
   left `ChatGPT.app` and `ChatGPT Classic.app` installed side by side instead of
   replacing it (matches @gilburns's observation in #3087 that the app renames itself
   in place on update — confirmed the same behavior below in the upgrade-path log).
   Added a small cleanup step, gated behind `DEBUG -eq 0` (same pattern as
   `microsoftcompanyportal.sh` etc.) so dry runs stay non-destructive, and guarded to
   only fire once (before the new bundle exists) so it's a one-time migration step.

Also aliased `chatgpt` and `chatgptclassic` to the same case-arm (two-line `|\`
continuation form, matching the existing convention in `adobereaderdc.sh` — needed so
Installomator's own label-listing regex still picks up both names) rather than only
renaming to `chatgptclassic`, so existing MDM deployments calling `chatgpt` keep
working while also giving @Rapha60 the exact `chatgptclassic` name requested in #3132.

Tested on a real Mac, both scenarios:
- Fresh install (no prior ChatGPT app): correct install path/bundle name
  (`/Applications/ChatGPT Classic.app`), version `1.2026.184`, Team ID `2DC432GLL2`
  matches, idempotent re-run reports no update needed, `chatgptclassic` alias resolves
  against the same install.
- Upgrade from an existing pre-184 `ChatGPT.app`: installs `ChatGPT Classic.app`
  alongside it rather than replacing it — Installomator has no way to know the old
  bundle is this app's predecessor. Per discussion below, cleaning up the old bundle
  is intentionally left to the deploying org's own workflow rather than handled by
  the label itself.

Fixes #3132

**Installomator log**

Upgrade-path run (the more interesting case — shows the legacy cleanup firing on line 10):

```
2026-08-04 10:55:50 : INFO  : chatgpt : setting variable from argument NOTIFY=silent
2026-08-04 10:55:50 : INFO  : chatgpt : setting variable from argument DEBUG=0
2026-08-04 10:55:50 : REQ   : chatgpt : ################## Start Installomator v. 10.10beta, date 2026-08-04
2026-08-04 10:55:50 : INFO  : chatgpt : ################## chatgpt
2026-08-04 10:55:50 : INFO  : chatgpt : Found legacy /Applications/ChatGPT.app from before the ChatGPT Classic rename - removing it so the label installs cleanly.
2026-08-04 10:55:51 : INFO  : chatgpt : Label type: dmg
2026-08-04 10:55:51 : INFO  : chatgpt : name: ChatGPT Classic, appName: ChatGPT Classic.app
2026-08-04 10:55:51 : WARN  : chatgpt : No previous app found
2026-08-04 10:55:51 : INFO  : chatgpt : Latest version of ChatGPT Classic is 1.2026.184
2026-08-04 10:55:51 : REQ   : chatgpt : Downloading https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg to ChatGPT Classic.dmg
2026-08-04 10:55:52 : INFO  : chatgpt : Downloaded ChatGPT Classic.dmg – Type is  zlib compressed data – Size is 82880 kB
2026-08-04 10:55:52 : REQ   : chatgpt : Installing ChatGPT Classic
2026-08-04 10:55:55 : INFO  : chatgpt : Mounted: /Volumes/ChatGPT Classic Installer
2026-08-04 10:55:56 : INFO  : chatgpt : Team ID matching: 2DC432GLL2 (expected: 2DC432GLL2 )
2026-08-04 10:55:56 : INFO  : chatgpt : Installing ChatGPT Classic version 1.2026.184 on versionKey CFBundleShortVersionString.
2026-08-04 10:55:56 : INFO  : chatgpt : Copy /Volumes/ChatGPT Classic Installer/ChatGPT Classic.app to /Applications
2026-08-04 10:55:59 : INFO  : chatgpt : found app at /Applications/ChatGPT Classic.app, version 1.2026.184, on versionKey CFBundleShortVersionString
2026-08-04 10:55:59 : REQ   : chatgpt : Installed ChatGPT Classic, version 1.2026.184
2026-08-04 10:55:59 : REQ   : chatgpt : All done!
2026-08-04 10:55:59 : REQ   : chatgpt : ################## End Installomator, exit code 0
```

### DIFF
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -2,6 +2,10 @@ chatgpt|\
 chatgptclassic)
     name="ChatGPT Classic"
     type="dmg"
+    if [[ -d "/Applications/ChatGPT.app" ]] && [[ ! -d "/Applications/ChatGPT Classic.app" ]] && [[ $DEBUG -eq 0 ]]; then
+        printlog "Found legacy /Applications/ChatGPT.app from before the ChatGPT Classic rename - removing it so the label installs cleanly." INFO
+        rm -rf "/Applications/ChatGPT.app"
+    fi
     if [[ $(arch) == arm64 ]]; then
         downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
     else
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -9,5 +9,4 @@ chatgptclassic)
     fi
     appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath 'string((//rss/channel/item/title)[1])' 2>/dev/null)
     expectedTeamID="2DC432GLL2"
-    blockingProcesses=( "ChatGPT Classic" )
     ;;
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -1,13 +1,14 @@
 chatgpt|\
 chatgptclassic)
     name="ChatGPT Classic"
-    type="dmg"
+    type="pkg"
     if [[ $(arch) == "arm64" ]]; then
-        downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
+        downloadURL="https://persistent.oaistatic.com/sidekick/public/ChatGPT_Classic.pkg"
     else
         printlog "ChatGPT Classic is only compatible with Apple Silicon (arm64) Macs." ERROR
         cleanupAndExit 95 "ChatGPT Classic requires Apple Silicon" ERROR
     fi
     appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath 'string((//rss/channel/item/title)[1])' 2>/dev/null)
     expectedTeamID="2DC432GLL2"
+    blockingProcesses=( "ChatGPT Classic" "ChatGPT" )
     ;;
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -2,15 +2,12 @@ chatgpt|\
 chatgptclassic)
     name="ChatGPT Classic"
     type="dmg"
-    if [[ -d "/Applications/ChatGPT.app" ]] && [[ ! -d "/Applications/ChatGPT Classic.app" ]] && [[ $DEBUG -eq 0 ]]; then
-        printlog "Found legacy /Applications/ChatGPT.app from before the ChatGPT Classic rename - removing it so the label installs cleanly." INFO
-        rm -rf "/Applications/ChatGPT.app"
-    fi
-    if [[ $(arch) == arm64 ]]; then
+    if [[ $(arch) == "arm64" ]]; then
         downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
     else
-        cleanupandexit 2 "No Intel-compatible download URL found. $appLabel is not Intel-compatible. Could not install app." ERROR
+        printlog "ChatGPT Classic is only compatible with Apple Silicon (arm64) Macs." ERROR
+        cleanupAndExit 95 "ChatGPT Classic requires Apple Silicon" ERROR
     fi
-    appNewVersion="$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath '(//rss/channel/item/title)[1]/text()' 2>/dev/null)"
+    appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath -q -e 'string((//rss/channel/item/title)[1])' 2>/dev/null)
     expectedTeamID="2DC432GLL2"
     ;;
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -8,6 +8,6 @@ chatgptclassic)
         printlog "ChatGPT Classic is only compatible with Apple Silicon (arm64) Macs." ERROR
         cleanupAndExit 95 "ChatGPT Classic requires Apple Silicon" ERROR
     fi
-    appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath -q -e 'string((//rss/channel/item/title)[1])' 2>/dev/null)
+    appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath 'string((//rss/channel/item/title)[1])' 2>/dev/null)
     expectedTeamID="2DC432GLL2"
     ;;
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -9,5 +9,5 @@ chatgptclassic)
     fi
     appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath 'string((//rss/channel/item/title)[1])' 2>/dev/null)
     expectedTeamID="2DC432GLL2"
-    blockingProcesses=( "ChatGPT Classic" "ChatGPT" )
+    blockingProcesses=( "ChatGPT Classic" )
     ;;
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -1,4 +1,3 @@
-chatgpt|\
 chatgptclassic)
     name="ChatGPT Classic"
     type="pkg"
--- a/fragments/labels/chatgpt.sh
+++ b/fragments/labels/chatgpt.sh
@@ -1,8 +1,9 @@
-chatgpt)
-    name="ChatGPT"
+chatgpt|\
+chatgptclassic)
+    name="ChatGPT Classic"
     type="dmg"
     if [[ $(arch) == arm64 ]]; then
-        downloadURL="https://persistent.oaistatic.com/sidekick/public/ChatGPT_Desktop_public_latest.dmg"
+        downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
     else
         cleanupandexit 2 "No Intel-compatible download URL found. $appLabel is not Intel-compatible. Could not install app." ERROR
     fi


gilburns:
Missing the appNewVersion. 

This URL seems to be what it is currently checking for it's own Sparkle based updates:
`https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml`

Hac3rsp4ce:
Thanks for looking! I think this might be GitHub's diff view cutting off context rather than the line actually being gone — `appNewVersion` is one line below where the hunk ends (right after the `fi`), unchanged from before:

```bash
appNewVersion="$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath '(//rss/channel/item/title)[1]/text()' 2>/dev/null)"
```

Same URL you mentioned. Confirmed it resolves correctly on a real Mac too — `Latest version of ChatGPT Classic is 1.2026.184` in the test log above. Let me know if you meant something else, though — happy to dig further if I'm missing what you're pointing at.

gilburns:
Yes, I see it now. My bad. 😁

uurazzle (changes requested):
Hi @Hac3rsp4ce:

Thanks for submitting the PR.

I’d suggest tightening it a bit so it stays closer to normal Installomator label behavior.

The main changes I’d recommend are:

- Remove the rm -rf "/Applications/ChatGPT.app" block. Labels generally shouldn’t delete an existing app bundle as part of version/download logic, especially with a hardcoded path.
- Fix cleanupandexit to cleanupAndExit; the current casing would fail.
- Keep the Apple Silicon check, but use Installomator’s standard unsupported-architecture pattern with printlog and exit code 95.
- Keep appNewVersion from the OpenAI appcast title, which currently matches the downloaded app’s CFBundleShortVersionString.
- Keep the label compact and avoid extra migration/removal behavior inside the label itself.


Here is a suggested label you can review and test.

```
chatgpt|chatgptclassic)
    name="ChatGPT Classic"
    type="dmg"
        downloadURL="https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg"
    else
        printlog "ChatGPT Classic is only compatible with Apple Silicon (arm64) Macs." ERROR
        cleanupAndExit 95 "ChatGPT Classic requires Apple Silicon" ERROR
    fi
    appNewVersion=$(curl -fs "https://persistent.oaistatic.com/sidekick/public/sparkle_public_appcast.xml" | xpath -q -e 'string((//rss/channel/item/title)[1])' 2>/dev/null)
    expectedTeamID="2DC432GLL2"
    ;;
```

### Verification Summary

- Official download: https://persistent.oaistatic.com/classic/public/ChatGPT_Classic.dmg
- Effective/final URL: same as downloadURL
- Resolved appNewVersion: 1.2026.184
- Payload type: DMG, mounted successfully and contained ChatGPT Classic.app
- Installed app version key: CFBundleShortVersionString=1.2026.184
- Bundle version: CFBundleVersion=1784145287
- Bundle identifier: com.openai.chat
- Architecture: app executable is arm64 only
- Gatekeeper/signing check: accepted, notarized Developer ID
- Verified Team ID: 2DC432GLL2

Hac3rsp4ce:
Thanks for this — really appreciate you actually testing it independently. Taking the `cleanupAndExit` casing fix and the standard arch-check pattern, those are clear improvements (and that casing bug predates this PR, good catch).

On dropping the `rm -rf` block — one thing I want to flag before I remove it: on a machine that already has the old pre-184 `ChatGPT.app` installed, without that cleanup step this label will install `ChatGPT Classic.app` right alongside it instead of replacing it, since Installomator's built-in detection has no way to know the old bundle is this app's predecessor. I confirmed this concretely on a real Mac (reproduced the duplicate before adding the cleanup, confirmed it goes away with it). Is that a known/accepted tradeoff for renamed-app labels in general, or is there a different pattern you'd use for this case? Happy to just drop it if that's the convention — didn't want to remove it silently without mentioning what it was covering.

uurazzle (changes requested):
Hi @Hac3rsp4ce :

Thanks for confirming the behavior. In this case, removal of older or differently named versions of the software should be handled separately by the organization’s MDM workflow or by a script that runs before or after Installomator.

Installomator labels should generally remain focused on identifying, downloading, validating, and installing the current application. Adding migration or removal logic for previous application names can introduce unexpected behavior and may delete an app that an administrator intended to retain.

I would recommend removing the rm -rf "/Applications/ChatGPT.app" block from the label and documenting that administrators should remove the older ChatGPT.app separately when deploying ChatGPT Classic.

Also, please remove the unnecessary line-continuation backslash from the label aliases and keep them on one line:

```
chatgpt|\
chatgptclassic)
```

vs

```
chatgpt|chatgptclassic)
```

Hac3rsp4ce:
Makes sense, thanks for the explanation — pushed a commit that removes the `rm -rf` block along with the other fixes (`cleanupAndExit` casing, standard arch-check pattern, your `xpath` invocation).

One small deviation from your snippet: I kept the alias as two lines (`chatgpt|\` / `chatgptclassic)`) rather than inline on one line — Installomator's own label-listing regex (used for `Labels.txt` and the no-arg output) only recognizes multi-name aliases when each name is on its own line ending in `|\` (matches the existing convention in `adobereaderdc.sh`); an inline `|` would make both names silently invisible from that listing even though the label itself still works. Wanted to flag that explicitly rather than just quietly changing what you suggested.

Should the duplicate-app limitation (old `ChatGPT.app` sticking around when upgrading from a pre-184 install) get documented anywhere in particular, or is mentioning it here in the PR thread enough?

uurazzle (dismissed):
Hi @Hac3rsp4ce :

Yes, this is accurate — and I checked it against the actual source rather than taking it on faith.

The regex Installomator uses to build Labels.txt and the no-argument label listing is (Installomator.sh, line 1429):

```
grep -E '^[a-z0-9\_-]*(\)|\|\\)$' "$0" | tr -d ')|\' | grep -v -E '^(broken.*|longversion|version|valuesfromarguments)$' | sort
```

That pattern only matches a line that is entirely a label name followed by either ) or |\ — the character class has no | in it, so a line like chatgpt|chatgptclassic) never matches at all (the pipe in the middle breaks it), and neither name would show up in Labels.txt or in ./Installomator.sh with no arguments. I confirmed this by running the actual regex against 
both forms:

`chatgpt|\ + chatgptclassic)` on separate lines → matches, both names extracted.
`chatgpt|chatgptclassic)` inline → matches nothing.

And adobereaderdc.sh does use exactly the multi-line convention you cited:

```
adobereaderdc|\
adobereaderdc-install|\
adobereaderdc-update)
```

So your correction is right on all counts: the label itself would still work fine (bash's case doesn't care about line breaks), but only the listing/Labels.txt generation would silently drop both alias names if written inline. Good catch, and worth flagging explicitly exactly as you did.

> Should the duplicate-app limitation (old ChatGPT.app sticking around when upgrading from a pre-184 install) get documented anywhere in particular, or is mentioning it here in the PR thread enough?

Yes, I think noting it here in the PR thread is good enough.

Hac3rsp4ce:
Updated the PR description to reflect the current behavior on the upgrade path (installs alongside the old app rather than replacing it, per our discussion above).

One more thing: the "Installomator log" in the description is from before the last commit — it still shows the `Found legacy /Applications/ChatGPT.app...` line from the cleanup step that's since been removed, so it no longer matches what the current code actually does. Would you rather I re-test on a Mac and swap in a fresh log against the current commit, or is a note above the existing log clarifying it's from an earlier revision good enough?

gilburns (changes requested):
The `xpath` does not need and should not include `-q -e`. There is a function in Installomator to insure that xpath works correctly no matter the version of macOS that it runs on.

Hac3rsp4ce:
Good catch — confirmed in `functions.sh`, the `xpath()` wrapper already adds `-q -e` (or just `-q` on older macOS) automatically, so passing them manually was duplicating/conflicting with that. Pushed a fix dropping the manual flags, kept the `string(...)` query as-is otherwise.

Hac3rsp4ce:
Given what you found in the `.pkg` script, we're switching this PR from `.dmg` to `.pkg` — same download family (sidekick bucket), same conventions we already settled on here (alias form, arch-check pattern, `xpath` call), plus `blockingProcesses` per lennart-duncker's note. The only real change is the package type and URL, which means the vendor's own `postinstall` migration handles the old `ChatGPT.app` → `ChatGPT Classic.app` swap instead of Installomator needing to.

Going to re-verify the actual upgrade path (pre-184 `ChatGPT.app` present, then update via the `.pkg`) on a real Mac before pushing, rather than relying only on the script read and the dry-run log, and will commit the change to this PR with fresh logs once that's confirmed. Flag anything before then if you see a reason to stick with `.dmg`.

Hac3rsp4ce:
Pushed the `.pkg` switch, plus one more fix a code review surfaced: `blockingProcesses` only listed `"ChatGPT Classic"`, which never matches on the actual migration path — the process running *before* the upgrade is named `ChatGPT`. Without catching that, the vendor's `postinstall` (`mv` old bundle aside, then `rm -rf` the backup) can leave a still-running old process orphaned with its files deleted, while a fresh instance opens alongside it. Added `"ChatGPT"` to the array to fix that.

@uurazzle — one open question rather than something I want to decide alone: `ChatGPT` is *also* the process name of OpenAI's separate new unified app (different from Classic, unrelated bucket). Adding it means Installomator could think that unrelated app blocks this install if it happens to be running. From reading the vendor script, that scenario looks like it'd fail at the postinstall level anyway (it explicitly checks bundle ID and refuses to run against the wrong app), so the practical risk seems low — but wanted your take given you've looked at this script more closely than I have. Keep `"ChatGPT"` in there, or is there a better way to scope this?

Logs from the real Mac, upgrade path (pre-184 `ChatGPT.app` present, then run against the new `.pkg` label):

```
2026-08-05 15:50:21 : INFO  : chatgpt : Label type: pkg
2026-08-05 15:50:21 : INFO  : chatgpt : archiveName: ChatGPT Classic.pkg
2026-08-05 15:50:21 : INFO  : chatgpt : name: ChatGPT Classic, appName: ChatGPT Classic.app
2026-08-05 15:50:21 : WARN  : chatgpt : No previous app found
2026-08-05 15:50:21 : INFO  : chatgpt : Latest version of ChatGPT Classic is 1.2026.184
2026-08-05 15:50:21 : REQ   : chatgpt : Downloading https://persistent.oaistatic.com/sidekick/public/ChatGPT_Classic.pkg to ChatGPT Classic.pkg
2026-08-05 15:50:23 : INFO  : chatgpt : Downloaded ChatGPT Classic.pkg – Type is  xar archive compressed TOC – Size is 82832 kB
2026-08-05 15:50:23 : REQ   : chatgpt : Installing ChatGPT Classic
2026-08-05 15:50:23 : INFO  : chatgpt : Team ID: 2DC432GLL2 (expected: 2DC432GLL2 )
2026-08-05 15:50:23 : INFO  : chatgpt : Installing ChatGPT Classic.pkg to /
2026-08-05 15:50:30 : INFO  : chatgpt : found app at /Applications/ChatGPT Classic.app, version 1.2026.184, on versionKey CFBundleShortVersionString
2026-08-05 15:50:30 : REQ   : chatgpt : Installed ChatGPT Classic, version 1.2026.184
2026-08-05 15:50:30 : REQ   : chatgpt : ################## End Installomator, exit code 0
```

`find /Applications -maxdepth 1 -iname "*chatgpt*"` right after that run — only one app remains, old bundle is gone:

```
/Applications/ChatGPT Classic.app
```

And the `chatgptclassic` alias, run afterward against the now-installed app — correctly detects it's already up to date instead of reinstalling:

```
2026-08-05 15:52:42 : INFO  : chatgptclassic : App(s) found: /Applications/ChatGPT Classic.app
2026-08-05 15:52:42 : INFO  : chatgptclassic : found app at /Applications/ChatGPT Classic.app, version 1.2026.184, on versionKey CFBundleShortVersionString
2026-08-05 15:52:42 : INFO  : chatgptclassic : appversion: 1.2026.184
2026-08-05 15:52:42 : INFO  : chatgptclassic : Latest version of ChatGPT Classic is 1.2026.184
2026-08-05 15:52:42 : INFO  : chatgptclassic : There is no newer version available.
2026-08-05 15:52:42 : REQ   : chatgptclassic : No newer version.
2026-08-05 15:52:42 : REQ   : chatgptclassic : ################## End Installomator, exit code 0
```

uurazzle (dismissed):
Hi @Hac3rsp4ce,

Thanks for contributing this Installomator label ([#3151: Fix chatgpt label: repoint stale download URL, alias chatgptclassic, clean up legacy app on upgrade](https://github.com/Installomator/Installomator/pull/3151))—everything looks good. Appreciate you taking the time to add it to the project!

Validation performed:
- Label(s) reviewed: `chatgpt`
- Required Installomator variables checked: `name`, `type`, `downloadURL`, `expectedTeamID`
- Recommended version handling checked: `appNewVersion`, standard app version lookup, and `versionKey`/custom version usage where applicable
- Built-in Installomator helpers/variables reviewed where applicable: `getJSONValue`, `xpath`, `archiveName`, `appName`, `packageID`, `appCustomVersion`, `downloadURLFromGit`, `versionFromGit`, and standard `$(arch)` handling
- Source and hygiene checks reviewed: direct vendor source, Homebrew Cask avoidance, jq avoidance, backslash continuations, commented lines, EditorConfig, PR template, and TeamID changes

Validation output summary:
- `chatgptclassic`: `Installomator/utils/assemble.sh chatgptclassic DEBUG=1` returned exit 0; debug execution `Yes`; URL reachable `Yes`; appNewVersion `1.2026.184`; TeamID `2DC432GLL2`; architecture `Unknown`

gilburns:
Throwing this out for consideration…

Should this label drop `chatgpt` and let that be assumed by the new ChatGPT (formerly Codex)?

```
chatgpt|\
chatgptclassic)
```

to 

`chatgptclassic)`

Hac3rsp4ce:
@gilburns Good question — rather than drop `chatgpt` now, what about following the same pattern used for `googlechrome` → `googlechromepkg`? That label was kept working as-is with a `printlog ... REQ` warning pointing admins at the preferred replacement, rather than an abrupt rename/removal.

Concretely: keep `chatgpt|\ chatgptclassic)` as it is now (so nothing currently deploying `chatgpt` breaks without notice), but add a warning like:
```bash
printlog "NOTE: The chatgpt label currently installs ChatGPT Classic. This name may be reassigned to OpenAI's new unified ChatGPT app in the future - use chatgptclassic explicitly if you want to guarantee installing ChatGPT Classic going forward." REQ
```
Once there's an actual label for the new unified app under the `chatgpt` name (if that's in the works separately from this PR), the alias here could be dropped on a real timeline instead of doing it speculatively today — same as how long Chrome's transition has been left running.

Open to just dropping it now if that's your/uurazzle's preference instead — wanted to raise the precedent and put a concrete option in front of you before deciding either way.

uurazzle:
Appreciate you laying out the concrete option, @gilburns. I'd still favor keeping it simple: chatgpt|\ chatgptclassic) with no added printlog.

The rename itself doesn't really need a separate notice, since every run already shows name: ChatGPT Classic and installs or finds ChatGPT Classic.app, so admins see the current app identity in the log regardless of any added message.

The "may be reassigned" half feels premature too — there's no unified-app label in the tree to actually collide with chatgpt yet that I have found, so I'd rather add a real, specific notice (or drop the alias outright) once that label exists, than carry a vague warning about a change that isn't happening yet.

gilburns:
The OpenAI Codex app is now named ChatGPT. This one is no longer named ChatGPT. That my reasoning for suggesting we transition the label that matches it to the other install. It's a mess that OpenAI created. 😁

Hac3rsp4ce:
@gilburns You're right and I checked — downloaded the current `codex-app-prod` artifact (`ChatGPT-darwin-arm64-26.803.41515.zip`, off `codex-app-prod/appcast.xml`) and inspected `Info.plist` directly:

```
CFBundleName        = ChatGPT
CFBundleDisplayName = ChatGPT
CFBundleExecutable  = ChatGPT
CFBundleIdentifier  = com.openai.codex   (unchanged)
```

The app really is named "ChatGPT" today, not a future possibility — `com.openai.codex` kept its bundle ID but the display name/executable are now literally "ChatGPT". `codex.sh` (`name="Codex"`) just hasn't caught up to that yet.

@uurazzle — this changes the premise from the earlier thread: there *is* now a real app using the "ChatGPT" name, it's just that the label matching it (`codex.sh`, PR #3087) hasn't been updated to reflect the rename. Given that, does this change your read on dropping `chatgpt` from this label's alias (leaving `chatgptclassic)` only), or would you rather that wait until `codex.sh`/#3087 actually gets fixed to claim the name? Happy to go either way — just didn't want to pick a side on a naming decision unilaterally.

gilburns:
The suggestion on this PR for Codex was to add chatgpt to the case statement:
https://github.com/Installomator/Installomator/pull/3087

It cannot exist in both labels. 😁

Hobadee:
My $0.02 as a sysadmin who needs to deploy both:
Give us a time-frame to switch labels over.  (3-6 months perhaps?)  During this time, add log output informing sysadmins of the impending change, and note it in release notes.  After this time, swap the `chatgpt` label from the `chatgptclassic` app to the `codex` app.

uurazzle:
Hi @Hac3rsp4ce:

Thanks for working through this. I think the `.pkg` approach is the right direction for the Classic label, especially because the vendor package should handle the old `ChatGPT.app` → `ChatGPT Classic.app` migration more safely than custom label logic.

One thing I’d like to verify before merging: since this is now a `.pkg` installer, can you check whether the package installs a stable receipt that uniquely identifies **ChatGPT Classic**?

If it does, adding `packageID` may be safer here than relying only on the app bundle name:

```bash
packageID="<verified ChatGPT Classic package id>"
```
The reason is that Installomator checks packageID first in getAppVersion. If the receipt exists, it uses the receipt version as the installed version before falling back to app bundle detection. That can be useful here because OpenAI has already renamed/migrated this app once, and future bundle-name changes could make app-path detection less reliable.

Before adding it, please verify:

- The receipt ID is specific to ChatGPT Classic.
- The receipt version matches the version returned by appNewVersion.
- The receipt remains stable across updates.
- It does not conflict with the newer unified ChatGPT/Codex app receipt, if that app installs one.

Also, I recommend making this PR explicitly Classic-only by using only:

```
chatgptclassic)
```
and not keeping chatgpt| as an alias. The chatgpt label should be reserved for the current unified ChatGPT app, while this label installs ChatGPT Classic.

So my recommended path is:

- Keep the .pkg installer.
- Change the case label to chatgptclassic) only.
- Verify whether a stable Classic-specific packageID exists.
- Add packageID if the receipt/version checks line up.
- Keep the Apple Silicon guard and the current appcast version lookup.

Thanks!

Hac3rsp4ce:
Hi @uurazzle — checked both points.

**On the packageID:** downloaded the actual `.pkg` and expanded it (`pkgutil --expand`, no install needed):

```
identifier="com.openai.chat.classic-update"
version="1784145287"
```

The identifier is specific to Classic, so that part checks out. But the version is the raw `CFBundleVersion` build number, not the marketing version — `appNewVersion` in the label resolves to `1.2026.184` (`CFBundleShortVersionString`, from the Sparkle appcast title). Those two will never compare equal. If I add `packageID` as-is, `getAppVersion` would read `1784145287` from the receipt, compare it against `1.2026.184`, always conclude there's a newer version, and reinstall on every run.

The current app-bundle detection (`versionKey=CFBundleShortVersionString`) already works correctly here — the PR's test log shows it correctly detecting "up to date" on a second run. Happy to add it if there's a way to reconcile the version formats I'm not seeing — let me know if you had something else in mind.

**On dropping the `chatgpt` alias:** done — switched the label to `chatgptclassic)` only.

gilburns:
@uurazzle 
I would argue against including the `packageID` in the label. It is generally only necessary for items that do not install into `/Applications`. 😁 

Here's the info about `packageID` from the Installomator wiki:

https://github.com/Installomator/Installomator/wiki/Label-Variables-Reference#strongly-recommended-variables

_`packageID` (optional, only recommended for non-app installs) This variable is for pkg bundle IDs. Do not use this variable if the app is in `/Applications`. Used for pkgs that only installs command line tools, which Installomator won't discover, or the app is not located in /Applications. (See label `desktoppr`, `golang`, `installomator_st`, `odrive`, or `teamviewerhost`) Keep in mind, if the app is uninstalled but the package receipt is still in place, Installomator will think it's installed._

gilburns:
> My $0.02 as a sysadmin who needs to deploy both: Give us a time-frame to switch labels over. (3-6 months perhaps?) During this time, add log output informing sysadmins of the impending change, and note it in release notes. After this time, swap the `chatgpt` label from the `chatgptclassic` app to the `codex` app.

@Hobadee - While I understand your caution here, the folks at OpenAI already exercised none of it in their haste to rename and rollout the app changes. I feel we need to at least keep up and try not to add to the confusion, by keeping the chatgpt label attached to the now wrong app label.

gilburns:
@Hac3rsp4ce - From a functionality perspective, I think this is good. Only one minor adjustment left. 

Blocking processes should remove "ChatGPT" from the list, since that is now associated with the former "Codex" app.

Hac3rsp4ce:
@gilburns Fixed — removed "ChatGPT" from `blockingProcesses`, keeping only "ChatGPT Classic".

uurazzle (changes requested):
Hi @Amaan:

Thanks for the PR.

### Remove redundant blockingProcesses app name

We noticed the label includes `blockingProcess`/`blockingProcesses` with only the app name.

For standard `/Applications` installs, Installomator can usually infer and handle the app process automatically, so explicitly setting the app name in `blockingProcesses` is often redundant.

For example, you have `name`:

```
name="ChatGPT Classic"
```

And `blockingProcesses`:

```
blockingProcesses=( "ChatGPT Classic" )
```

Please remove unnecessary `blockingProcess`/`blockingProcesses` app-name entries unless there is a specific non-default process that really needs to be handled. Thanks!

gilburns (dismissed):

Hac3rsp4ce:
@uurazzle Fixed — dropped `blockingProcesses`. Confirmed on a real Mac: Installomator infers it from `name` automatically ("no blocking processes defined, using ChatGPT Classic as default"), and a full end-to-end install still succeeds:

```
2026-08-24 15:04:10 : INFO  : chatgptclassic : no blocking processes defined, using ChatGPT Classic as default
2026-08-24 15:04:10 : INFO  : chatgptclassic : name: ChatGPT Classic, appName: ChatGPT Classic.app
2026-08-24 15:04:11 : REQ   : chatgptclassic : Downloading https://persistent.oaistatic.com/sidekick/public/ChatGPT_Classic.pkg to ChatGPT Classic.pkg
2026-08-24 15:04:11 : REQ   : chatgptclassic : no more blocking processes, continue with update
2026-08-24 15:04:11 : REQ   : chatgptclassic : Installing ChatGPT Classic
2026-08-24 15:04:17 : INFO  : chatgptclassic : Finishing...
2026-08-24 15:04:20 : INFO  : chatgptclassic : App(s) found: /Applications/ChatGPT Classic.app
2026-08-24 15:04:20 : REQ   : chatgptclassic : Installed ChatGPT Classic, version 1.2026.184
2026-08-24 15:04:20 : REQ   : chatgptclassic : All done!
2026-08-24 15:04:20 : REQ   : chatgptclassic : ################## End Installomator, exit code 0
```

uurazzle (approved):
Hi [Hac3rsp4ce](https://github.com/Hac3rsp4ce),

Thanks for contributing this Installomator label ([#3151: Fix chatgpt label: repoint stale download URL, alias chatgptclassic, clean up legacy app on upgrade](https://github.com/Installomator/Installomator/pull/3151))—everything looks good. Appreciate you taking the time to add it to the project!

Validation performed:
- Label(s) reviewed: `chatgpt`
- Required Installomator variables checked: `name`, `type`, `downloadURL`, `expectedTeamID`
- Recommended version handling checked: `appNewVersion`, standard app version lookup, and `versionKey`/custom version usage where applicable
- Built-in Installomator helpers/variables reviewed where applicable: `getJSONValue`, `xpath`, `archiveName`, `appName`, `packageID`, `appCustomVersion`, `downloadURLFromGit`, `versionFromGit`, and standard `$(arch)` handling
- Source and hygiene checks reviewed: direct vendor source, Homebrew Cask avoidance, jq avoidance, backslash continuations, commented lines, EditorConfig, PR template, and TeamID changes

Validation output summary:
- `chatgptclassic`: `Installomator/utils/assemble.sh chatgptclassic DEBUG=1` returned exit 0; debug execution `Yes`; URL reachable `Yes`; appNewVersion `1.2026.184`; TeamID `2DC432GLL2`; architecture `Unknown`

gilburns (approved):

gilburns:
@uurazzle @dan-snelson 

Probably we hold off merging this until the Codex/ChatGPT label is fully approved and merged at the same time?

https://github.com/Installomator/Installomator/pull/3087

gilburns:
Or not 😁

uurazzle:
Sorry, too quick. I apologize.

gilburns:
@Hac3rsp4ce - Would you mind opening up another PR that simply renames the sh file?

cd in the labels directory and the do this:
`git mv chatgpt.sh chatgptclassic.sh`

The push that PR into here for approval?

Hac3rsp4ce:
@gilburns Opened the rename PR you asked for: https://github.com/Installomator/Installomator/pull/3183

PR #3166, opened 2026-08-11
fix(nudgesuite): detect suite via packageID receipt instead of app version

### LINKED ISSUE #3038: Broken app label: nudgesuite
**What's wrong with the app label?**<br>There is no effective difference between "nudge" and "nudgesuite." Both install Nudge.app, however Nudge Suite includes the logger and the LaunchAgent. If Nudge.app is already installed but the rest of the suite is not, the nudgesuite label doesn't do anything. Do distinguish it from the "nudge" label, it should detect not just Nudge.app but the logger and the LaunchAgent.

**Have you confirmed this Issue has not been reported already? (Check open Issues and Pull requests)**
Yes.

**Installomator version**<br>What version are you running? You must include the version number and date. (Ex: Installomator v. 10.9-beta, date 2026-06-04

**Additional context**<br>Note in the logging below, Nudge.app was installed, but /Library/LaunchAgents/com.github.macadmins.Nudge.plist and /Library/LaunchDaemons/com.github.macadmins.Nudge.logger.plist had been removed. They are the key difference between "nudge" and "nudgesuite".
The way we found this is that we deploy Nudge app updates from Jamf App Installers. However, they have only the core Nudge app, not the Nudge Suite package. So we found a number of machines did not have the Logger and the LaunchAgent, even though they had the Nudge core app. When we went to run Installomator with the "nudgesuite" label, it wasn't fixing them because its detection isn't specific enough.

**Installomator log**<br>
At the bottom of this report, provide a log, either by running Installomator in Terminal and saving the output, or by downloading the result log from your MDM. **Do not enable [Debug logging level](https://github.com/Installomator/Installomator/wiki/Configuration-and-Variables#logging-level) and please format the log [using a code block](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks)!**
Attached.

Hac3rsp4ce:
Yes.

Yes — only `fragments/labels/nudgesuite.sh`.

**Did you use our editorconfig file?**
Yes.

**Additional context**
Fixes the bug reported in #3038: `nudgesuite` shared app-version detection with the plain `nudge` label, so if `Nudge.app` was already installed (e.g. via `nudge`, or a Jamf App Installer shipping only the core app), `nudgesuite` saw the app at the current version and skipped installing entirely — the suite's LaunchAgent/LaunchDaemon files never got laid down.

Per @uurazzle's suggestion on the issue thread, switched detection to Installomator's `packageID` mechanism, keyed on the Suite pkg's own identifier (`com.github.macadmins.Nudge.Suite`), confirmed distinct from the plain pkg's `com.github.macadmins.Nudge`. `appName` is intentionally omitted so the fallback app-search can't accidentally match the already-installed `Nudge.app` by name and short-circuit the receipt check.

Tested on a real Mac: installed the core app via `nudge` alone, then ran the fixed `nudgesuite` — it correctly ignored the pre-existing app, installed the Suite pkg (Team ID `T4SK8ZXCXG` matches), and both `/Library/LaunchAgents/com.github.macadmins.Nudge.plist` and `/Library/LaunchDaemons/com.github.macadmins.Nudge.logger.plist` were present afterward. A second run correctly reported no update needed.

Fixes #3038

**Installomator log**
```
2026-08-11 12:19:07 : REQ   : nudgesuite : ################## Start Installomator v. 10.10beta, date 2026-08-11
2026-08-11 12:19:07 : DEBUG : nudgesuite : packageID=com.github.macadmins.Nudge.Suite
2026-08-11 12:19:07 : INFO  : nudgesuite : found packageID com.github.macadmins.Nudge.Suite installed, version 2.1.3.81860
2026-08-11 12:19:08 : DEBUG : nudgesuite : origin=Developer ID Installer: Mac Admins Open Source (T4SK8ZXCXG)
2026-08-11 12:19:08 : INFO  : nudgesuite : Team ID: T4SK8ZXCXG (expected: T4SK8ZXCXG )
2026-08-11 12:19:08 : INFO  : nudgesuite : Checking package version.
2026-08-11 12:19:08 : INFO  : nudgesuite : Downloaded package com.github.macadmins.Nudge.Suite version 2.1.3.81860
2026-08-11 12:19:08 : INFO  : nudgesuite : Downloaded version of Nudge Suite is the same as installed.
2026-08-11 12:19:08 : REQ   : nudgesuite : No new version to install
2026-08-11 12:19:08 : REQ   : nudgesuite : ################## End Installomator, exit code 0
```

### DIFF
--- a/fragments/labels/nudgesuite.sh
+++ b/fragments/labels/nudgesuite.sh
@@ -1,10 +1,10 @@
 nudgesuite)
     name="Nudge Suite"
-    appName="Nudge.app"
     type="pkg"
     archiveName="Nudge_Suite-[0-9.]*.pkg"
-    appNewVersion=$(versionFromGit macadmins Nudge )
+    packageID="com.github.macadmins.Nudge.Suite"
     downloadURL=$(downloadURLFromGit macadmins Nudge )
+    appNewVersion=$(versionFromGit macadmins Nudge )
     expectedTeamID="T4SK8ZXCXG"
     blockingProcesses=( "Nudge" )
     ;;


gilburns:
If you really want the detection to work, then the nudge.sh label would also need to be updated. It will see the "Nudge" app installed by the nudgesuite.sh label as being installed. Just saying. 😁

gilburns (approved):

uurazzle (changes requested):
Hi @Amaan:

Thanks for the PR.

### Use installed app name

This label can be simplified by setting `name` to the installed app bundle name in `/Applications`.

Installomator derives several defaults from `name`, including `appName="$name.app"` and `blockingProcesses=( "$name" )`. When the installed app is named differently than the label's display name, using the installed app name lets Installomator infer those defaults and avoids extra label code.

Please update the label as follows:

Detected in `nudgesuite`:

```bash
name="Nudge"
```

Remove any now-redundant `appName`, `blockingProcesses`, or `packageID` entries that only duplicate what Installomator can infer from that `name`.

And if you change the `name` you the `blockingProcesses` is unnecessary:

```
blockingProcesses=( "Nudge" )
```

Keep other variables such as `archiveName` when they are still needed to select the correct download asset. Thanks!

Hac3rsp4ce:
@uurazzle Thanks for the review!

Before I make this change — I think setting `name="Nudge"` here would actually reintroduce the exact bug #3038 reported, so I want to flag the specific mechanism before touching it:

Right now `name="Nudge Suite"` is left as-is *deliberately*, because `appName` isn't set explicitly. Per `fragments/main.sh:165-168`, when `appName` is unset it defaults to `"$name.app"` — so today that's `"Nudge Suite.app"`, a bundle that never actually exists on disk (the suite installs `Nudge.app`, just with extra LaunchAgent/LaunchDaemon files alongside it). That's intentional: it means the app-path fallback in `getAppVersion` can never accidentally match the plain `Nudge.app` that a standalone `nudge` install leaves behind.

Here's why that matters: `getAppVersion` (`fragments/functions.sh:229-249`) checks `packageID` first, and only falls through to the app-path search if there's no receipt for it. If we rename to `name="Nudge"`, `appName` would default to `"Nudge.app"` instead — which *does* exist whenever `nudge` was run previously. So in exactly the "suite not installed yet" case (no `com.github.macadmins.Nudge.Suite` receipt), it'd fall through, match the pre-existing `Nudge.app`, and report "up to date" — which is precisely the false-negative #3038 was filed about.

`packageID` also isn't derived from `name` anywhere in the code, so I don't think it's redundant here regardless — it's the actual detection mechanism this PR relies on.

Totally on board with simplifying where it's safe though. A couple options, happy to go with whichever you prefer:
1. Leave `name="Nudge Suite"` as-is (keeps the non-existent-bundle fallback intact), keep the explicit `blockingProcesses=("Nudge")` override since it can't be inferred while `name` stays "Nudge Suite".
2. Rename to `name="Nudge"` (drops the now-redundant `blockingProcesses` override) but add back an explicit `appName="Nudge Suite.app"` so the fallback still can't match the standalone app.

Let me know which you'd rather have, or if I'm missing something about how this should work.

Hac3rsp4ce:
@gilburns Good point, but I think that's a different issue from this one.

This PR only fixes `nudgesuite.sh`: it now correctly notices when the suite parts are missing, even if `Nudge.app` is already there from the plain `nudge` label. What you're describing is the opposite case — `nudge.sh` thinking it's "up to date" when really the full suite was installed. That's a real gap, but it's in a different file, and `nudge.sh` was never meant to know about the suite in the first place.

I didn't want to change two labels in one PR. Want me to open a separate issue for `nudge.sh`, or do you think it's fine to leave as is?

uurazzle:
Hi @Hac3rsp4ce:

Thanks for the explanation. I agree with your reasoning here.

For this label, it is better to detect the installed state of the **suite package**, not just the shared `Nudge.app` bundle?

With the current main labels:

```
nudge)
    name="Nudge"
    ...
    archiveName="Nudge-[0-9.]*.pkg"
```

and:

```
nudgesuite)
    name="Nudge Suite"
    appName="Nudge.app"
    ...
    archiveName="Nudge_Suite-[0-9.]*.pkg"
```

Both labels use the same installed app bundle for version detection:

```
/Applications/Nudge.app
```

So if `nudge` label updates Nudge.app to the latest version, then `nudgesuite` label can look at that same Nudge.app, see the latest version, and conclude “already installed/up to date,” even if the suite-only LaunchAgent/LaunchDaemon payloads are missing.

That means if the nudge label updates Nudge.app to the latest version, the nudgesuite label can also see that same app bundle, decide it is already current, and skip installation even when the suite-specific LaunchAgent/LaunchDaemon payloads are missing.

That is why packageID matters for nudgesuite: it detects the suite package receipt, not just the shared app bundle.

With this label:

```
packageID="com.github.macadmins.Nudge.Suite"
```

`getAppVersion` checks the pkg receipt first using pkgutil:

```
pkgutil --pkg-info-plist com.github.macadmins.Nudge.Suite
```

If that receipt exists and has a version, Installomator uses that as the installed version and sets:

```
updateDetected="YES"
```

Then it compares that installed version against:

```
appNewVersion=$(versionFromGit macadmins Nudge )
```

The suite package asset is still selected by:

```
archiveName="Nudge_Suite-[0-9.]*.pkg"
```

So packageID makes sense here. It avoids confusing “Nudge.app is current” with “the full Nudge Suite package is installed.”

Given that, I’m good with keeping `name="Nudge Suite"`, `packageID="com.github.macadmins.Nudge.Suite"`, and the explicit `blockingProcesses=( "Nudge" )`.

Hac3rsp4ce:
@uurazzle Thanks for confirming — that's exactly the mechanism, glad we're aligned. Since the `packageID` check is what actually resolves it, let me know if you're good to update your review to approved so this can move toward merging.

uurazzle (approved):
Hi [Hac3rsp4ce](https://github.com/Hac3rsp4ce),

Thanks for contributing this Installomator label ([#3166: fix(nudgesuite): detect suite via packageID receipt instead of app version](https://github.com/Installomator/Installomator/pull/3166))—everything looks good. Appreciate you taking the time to add it to the project!

Validation performed:
- Label(s) reviewed: `nudgesuite`
- Required Installomator variables checked: `name`, `type`, `downloadURL`, `expectedTeamID`
- Recommended version handling checked: `appNewVersion`, standard app version lookup, and `versionKey`/custom version usage where applicable
- Built-in Installomator helpers/variables reviewed where applicable: `getJSONValue`, `xpath`, `archiveName`, `appName`, `packageID`, `appCustomVersion`, `downloadURLFromGit`, `versionFromGit`, and standard `$(arch)` handling
- Source and hygiene checks reviewed: direct vendor source, Homebrew Cask avoidance, jq avoidance, backslash continuations, commented lines, EditorConfig, PR template, and TeamID changes

Validation output summary:
- `nudgesuite`: `Installomator/utils/assemble.sh nudgesuite DEBUG=1` returned exit 0; debug execution `Yes`; URL reachable `Yes`; appNewVersion `2.1.3.81860`; TeamID `T4SK8ZXCXG`; architecture `Universal`

PR #3194, opened 2026-08-28
When available, include "v$appNewVersion" in progresstext

HowardGMac:
---
Yes
No
Yes

This is a revival of PR #794 originally posted by Dan Snelson. Seems it was never incorporated into the main Installomator script, but would be a solid feature to have in the swiftDialog integration.

When available, include version number in 'progresstext' (but do not modify 'appNewVersion' variable)

### DIFF
--- a/Installomator.sh
+++ b/Installomator.sh
@@ -1400,7 +1400,12 @@ updateDialog() {
             echo "progress: $progress" >> $cmd_file
         fi
         if [[ $message != "" ]]; then
-            echo "progresstext: $message" >> $cmd_file
+            if [[ $appNewVersion != "" ]]; then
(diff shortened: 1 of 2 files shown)


acodega:
Changes should be made against the fragments, in functions.sh for example, not the script itself.

gilburns:
@HowardGMac - You want to make this change in the `functions.sh` file instead. Then when the `Installomator.sh` script is assembled, it would be reflected in there.

`./fragments/functions.sh`

HowardGMac:
My apologies, the notes about making contributions suddenly became crystal clear. Not sure how I missed that. Please let me know if I need to correct this further.

